PR #4, opened 2020-01-30
iOS: Monkey patch loading view to support iOS 13

tido64:


tido64:
| Before | After  |
|:------:|:------:|
 | image | image |

sveinar:
:drooling_face:

PR #8, opened 2020-03-02
iOS: Set up example project and build CI

tido64:


tido64:
@ArazAbishov: Yup, I thought of that as well. I was hoping to bring that along with the manifest work we were hoping to get done this week.

PR #9, opened 2020-03-03
iOS: App manifest prototype

tido64:
ReactTestApp


afoxman:
Hi Tommy,

I've got [code you can use](https://github.com/microsoft/continuity) for dealing with manifests. It's a new project, and I started with Windows first, but I have every intention of making it work for Apple and Android platforms, too.

Look at the [public headers](https://github.com/microsoft/continuity/blob/master/include/Continuity/ReactManifest.h) to see the manifest schema. It's based on some already-outdated ideas. In issues [4](https://github.com/microsoft/continuity/issues/4), [5](https://github.com/microsoft/continuity/issues/5) and [6](https://github.com/microsoft/continuity/issues/6), I've got notes on how it will change.

Our projects are likely to converge when the Apple platforms come online. Maybe you're available to get them running? :) If not, let's make sure we're staying in sync on schema ([which still needs to be formally documented](https://github.com/microsoft/continuity/issues/3)).

tido64:
@afoxman: I've created an issue for it. I'll check this in for now so we can start iterating on it.

PR #13, opened 2020-03-07
Configure clang-format and SwiftLint

tido64:


ios/ReactTestApp/AppDelegate.swift
```diff
@@ -1,9 +1,9 @@
-/**
- * Copyright (c) Microsoft Corporation. All rights reserved.
- *
- * This source code is licensed under the MIT license found in the
- * LICENSE file in the root directory of this source tree.
- */
+//
```
ArazAbishov: Do we need to add this header to all files? It will be cleaner just to have the license file in the repo root.

tido64: This is not the full license though. It just says where you can find it.

ArazAbishov: Correct me if I am wrong, but I believe that you don't have to link the license in every source code file. I would rather prefer not to have it, as it just adds noise to files.

tido64: It is not required but is recommended practice. If people were sent single files separated from the project, they would still know what license is applied. I’d prefer to leave it in.

PR #15, opened 2020-03-07
iOS: Add Example tests

tido64:


ArazAbishov:
This PR contains changes from another PR, I will wait for the former one to get merged first.

ios/ReactTestApp/Manifest.swift
```diff
@@ -7,21 +7,25 @@
...
 }
 
-struct Manifest: Decodable {
+struct Manifest: Codable {
     let name: String
     let displayName: String
     let components: [String: Component]
 
-    static func readFile() -> Manifest? {
+    static func fromFile() -> Manifest? {
```
ArazAbishov: Would it make sense to parameterise this function, or to provide an overloaded method that accepts a path / json? I am thinking of scenarios where manifest might be not in the project root, or when the user might have several manifests?

tido64: Let's add overloads when we'll actually need them.

PR #16, opened 2020-03-09
iOS: Use Node to resolve package paths

tido64:


test_app.rb
```diff
@@ -1,15 +1,27 @@
...
+#
 
-def use_test_app!(packageRoot)
+def autolink_script_path()
+  package_path = resolve_module('@react-native-community/cli-platform-ios')
+  return File.join(package_path, 'native_modules')
+end
+
+def resolve_module(request)
+  script = "console.log(path.dirname(require.resolve('#{request}/package.json')));"
```
alloy: Huh, TIL that `node` automatically imports `path` 👌

PR #17, opened 2020-03-09
feat: Bump react-native to 0.61.5

tido64:


example/react-native.config.js
```diff
@@ -5,6 +5,11 @@ if (process.argv.includes("--config=metro.config.macos.js")) {
 } else {
   module.exports = {
     project: {
+      android: {
+        sourceDir: "android",
+        manifestPath:
+          "../node_modules/react-native/template/android/app/src/main/AndroidManifest.xml",
```
ArazAbishov, line 11: Was this necessary? I would rather check-in the manifest file into example.

tido64: It's mostly to keep the `android` folder clean and as close to a real world package as possible:
```sh
% find android
android
android/gradle
android/gradle/wrapper
android/gradlew
android/build.gradle
android/gradle.properties
android/gradlew.bat
android/settings.gradle
```
Should I check it in?

tido64: Actually, it looks like I can do without? Don't recall why it was needed in the first place ...

ArazAbishov: When I was working with the example there was no need for the manifest. Removing it would be ideal or alternatively checking it in. Otherwise RN CLI tools will have a ref to a manifest while AS doesn't, which can cause problems when working with it from IDE.

tido64: It's removed.

PR #19, opened 2020-03-16
Add instructions for setting up dev env for iOS

tido64:


README.md
```diff
@@ -9,6 +9,65 @@ The package is still a bit cumbersome to consume, and requires manually writing
...
+We'll be using the Example app for all development of the React Native Test App.
+Some platforms may require extra steps for the initial set up. Please follow the
+steps below, then jump to the appropriate section(s) for the final steps.
+
+Open a terminal and navigate to your clone of this repository:
+
+```bash
+cd react-native-test-app
+
+# This step only needs to be done once. It is stored in
```
alloy: Perhaps an addition like this will help readers understand the order, I at first assumed you were missing the `yarn install` that most repos do at the very start.

```suggestion
# This step only needs to be done once, before we can install the Example app’s dependencies. It is stored in
```

PR #21, opened 2020-03-16
iOS: Improve syntax for providing targeted pods

tido64:


tido64:
@alloy: Would you say this is more idiomatic Ruby? Any better suggestions?

example/Podfile
```diff
@@ -2,12 +2,12 @@ require_relative 'node_modules/react-native-test-app/test_app.rb'
...
-end
+add_app_pods = Proc.new {
+  pod 'Example-dev', :path => '.'
+}
+
+add_test_pods = Proc.new {
+  pod 'Example-Tests', :path => '.'
+}
+
+use_test_app!(__dir__, :add_app_pods => add_app_pods, :add_test_pods => add_test_pods)
```
alloy: What do you think of something like the following?

```ruby
use_test_app!(__dir__) do |target|
  target.app do
    pod 'Example-dev', :path => '.'
  end

  target.test do
    pod 'Example-Tests', :path => '.'
  end
end
```

alloy: I feel like `use_test_app` could be improved upon too, but don’t have a concrete suggestion right now. Will sleep on it.

tido64: Definitely! Please feel free to make any ergonomic improvements.

I pushed some changes. Let me know if it's what you had in mind.

alloy: Yup, looks great 👌

PR #22, opened 2020-03-18
iOS: Install resources from app manifest

tido64:


tido64:
@alloy: How do you feel about generating a temporary resources podspec?

alloy:
It’s not the worst thing 🤔 I was wanting to see about generating an in-memory podspec, but this may actually be more API proof.

tido64:
I looked into it yesterday. Worked back from [Core/dsl.rb](https://github.com/CocoaPods/Core/blob/master/lib/cocoapods-core/podfile/dsl.rb), got as far as the [analyzer](https://github.com/CocoaPods/CocoaPods/blob/e5485d55addf291a61f1eae9bfc2f1fe1fe95b18/lib/cocoapods/installer/analyzer.rb#L101), and gave up. I couldn't find a proper way to inject `PodTargets`. Not sure about this solution myself.

test_app.rb
```diff
@@ -82,6 +108,9 @@ def use_test_app!(package_root)
     pod 'glog', :podspec => "#{react_native}/third-party-podspecs/glog.podspec"
     pod 'Folly', :podspec => "#{react_native}/third-party-podspecs/Folly.podspec"
 
+    resources_pod_path = resources_pod(package_root)
+    pod 'ReactTestApp-Resources', :path => resources_pod_path unless resources_pod_path.nil?
```
alloy:
```ruby
if resources_pod_path = resources_pod(package_root)
  pod 'ReactTestApp-Resources', :path => resources_pod_path
end
```

tido64:
@alloy: Should we go ahead with this for now and figure out a better implementation later? The underlying implementation should be invisible to consumers in any case.

PR #23, opened 2020-03-19
Add Plop template

tido64:


plopfile.js
```diff
@@ -0,0 +1,143 @@
...
+            );
+          }
+        }
+      ];
+
+      const exclusive = platforms !== "All";
+      if (!exclusive || platforms === "iOS") {
+        actions.push({
+          type: "add",
+          path: exclusive ? "Podfile" : "ios/Podfile",
```
tido64: This doesn't quite work unless we take a direct dependency on `@react-native-community/cli` 3.x, or simply bump React Native to 0.61. Still have to figure out whether we can simply do the former.

tido64: Took a dependency on `@react-native-community/cli` 4.x for now. It'll have to be removed when we bump React Native.

sweggersen:
LGTM

PR #24, opened 2020-03-20
Support for the example app on Android

ArazAbishov:
What is there:
 * Support for the following use-cases: 
   * library without native code and native dependencies
   * library without native code and with native dependencies
   * library with native code and without native dependencies
   * library with native code and with native dependencies
 * Manifest support
 * Packaging bundle

What is missing:
 * Support for native controllers. Test-app interprets all `app.json` entry points as JS components. 
 * Loading resources from `app.json`, rather than from `gradle.properties`

Workarounds:
 * Forked autolinking plugin into `test-app-native-modules.gradle`. It was necessary to force the plugin to generate linking code with imports for the `react-native-test-app`. The change introduced in the fork is "up-streamable", but needs to be discussed first.


android/app/build.gradle
```diff
@@ -1,38 +1,81 @@
...
+
+project.ext.react = [ enableHermes: true ]
 
 android {
     compileSdkVersion 29
     buildToolsVersion "29.0.3"
 
-    defaultConfig {
-        applicationId "com.sample"
+    defaultConfig {
```
ArazAbishov: Re-format the file.

android/app/build.gradle
```diff
@@ -1,38 +1,81 @@
...
-    implementation 'androidx.core:core-ktx:1.0.2'    
-    testImplementation 'junit:junit:4.12'
-    androidTestImplementation 'androidx.test.ext:junit:1.1.0'
-    androidTestImplementation 'androidx.test.espresso:espresso-core:3.1.1'
+
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlinVersion"
+    implementation "androidx.appcompat:appcompat:1.1.0"
+    implementation "androidx.core:core-ktx:1.2.0"
+    implementation "androidx.recyclerview:recyclerview:1.1.0"
+    implementation "androidx.swiperefreshlayout:swiperefreshlayout:1.0.0"
```
ArazAbishov: FYI: for some reason, RN crashes if this dependency is not on the classpath  🤷‍♂

android/app/src/main/java/com/sample/ComponentActivity.kt
```diff
@@ -0,0 +1,116 @@
...
+            .setInitialLifecycleState(LifecycleState.BEFORE_RESUME)
+            .addPackages(PackageList(application).packages)
+            .setApplication(application)
+            .setCurrentActivity(this)
+
+        val bundleName = findBundleFile()
+        if (bundleName == null) {
+            reactInstanceManagerBuilder
+                .setUseDeveloperSupport(BuildConfig.DEBUG)
+                .setJSMainModulePath("index")
```
ArazAbishov: A note for myself: check if iOS parameterises the module path name when fetching the bundle from dev server. cc @tido64

android/app/src/main/java/com/sample/ComponentActivity.kt
```diff
@@ -0,0 +1,116 @@
...
+
+        componentName?.let {
+            reactRootView.startReactApplication(
+                reactInstanceManager, it, null
+            )
+        }
+    }
+
+    private fun findBundleFile(): String? {
+        val possibleEntryFiles = listOf(
```
ArazAbishov: Extract into a constant.

android/app/src/main/java/com/sample/MainActivity.kt
```diff
@@ -1,77 +1,82 @@
...
-            .setBundleAssetName("index.android.bundle")
-            .setJSMainModulePath("index")
-            .setApplication(application)
-            .build()
-
-        reactRootView.startReactApplication(
-            reactInstanceManager, "TestComponent", null
-        )
-    }
+        val moshi = Moshi.Builder().build()
```
ArazAbishov: Extract Manifest parsing logic into a separate class.

android/app/src/main/java/com/sample/MainActivity.kt
```diff
@@ -1,77 +1,82 @@
...
+            adapter = ComponentAdapter(LayoutInflater.from(context), components)
 
-        reactInstanceManager.onHostPause(this)
+            addItemDecoration(DividerItemDecoration(context, DividerItemDecoration.VERTICAL))
+        }
     }
 
-    override fun onResume() {
-        super.onResume()
+    private inner class ComponentAdapter(
```
ArazAbishov: Extract into a separate file.

android/app/src/main/java/com/sample/MainActivity.kt
```diff
@@ -1,77 +1,82 @@
...
+import androidx.recyclerview.widget.DividerItemDecoration
+import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView
+import androidx.recyclerview.widget.RecyclerView.Adapter
+import androidx.recyclerview.widget.RecyclerView.ViewHolder
+import com.squareup.moshi.Moshi
+
+class MainActivity : AppCompatActivity() {
+
+    data class ComponentViewModel(val name: String, val displayName: String)
```
ArazAbishov: Extract into a separate file.

example/gradle.properties
```diff
@@ -0,0 +1,9 @@
+# This properties are required to enable 
+# AndroidX for the test-app.
+android.useAndroidX=true
+android.enableJetifier=true
+
+# A comma separated list of directories that are 
+# expected to be copied to the test app.
+testApp.bundle=dist/main.jsbundle
```
ArazAbishov: Create a task to add support for loading resources from `app.json` instead.

example/package.json
```diff
@@ -13,7 +14,7 @@
     "@babel/core": "^7.0.0",
     "mkdirp": "^0.5.1",
     "react": "16.8.6",
-    "react-native": "0.60.6",
+    "react-native": "0.61.5",
```
ArazAbishov: Avoid bumping to 61. Two possible solutions: 
1) In `test-app-native-modules.gradle`, use two different commands to find the path to CLI binaries (see iOS's autolinking implementation)
2) Bump CLI tools independently.

package.json
```diff
@@ -24,7 +24,7 @@
     "@react-native-community/cli-platform-ios": "^2.4.1",
     "metro-react-native-babel-preset": "0.54.1",
     "react": "16.8.6",
-    "react-native": "0.60.6"
+    "react-native": "0.61.5"
```
ArazAbishov: See the comment above.

test-app.gradle
```diff
@@ -1,176 +1,56 @@
...
-def plugin = new TestAppPlugin(
-    generatedFileName, 
-    generatedFilePackage, 
-    generatedFileContentsTemplate
-)
-
-ext.applyTestAppSettings = { DefaultSettings defaultSettings -> 
-    plugin.apply(defaultSettings) 
-}
+    task copyResources(type: Copy) {
```
ArazAbishov: Create a task to add support for copying resources from `app.json`.

android/app/build.gradle
```diff
@@ -1,38 +1,81 @@
...
+project.ext.react = [ enableHermes: true ]
 
 android {
     compileSdkVersion 29
     buildToolsVersion "29.0.3"
 
-    defaultConfig {
-        applicationId "com.sample"
+    defaultConfig {        
+        applicationId "com.sample"
```
tido64: Trim
```suggestion
    defaultConfig {
        applicationId "com.sample"
```

test-app-native-modules.gradle
```diff
@@ -0,0 +1,271 @@
...
+
+ext.applyNativeModulesSettingsGradle = { DefaultSettings defaultSettings, String root = null ->
+  if (root != null) {
+    logger.warn("${ReactNativeModules.LOG_PREFIX}Passing custom root is deprecated. CLI detects root automatically now.");
+    logger.warn("${ReactNativeModules.LOG_PREFIX}Please remove second argument to `applyNativeModulesSettingsGradle`.");
+  }
+  autoModules.addReactNativeModuleProjects(defaultSettings)
+}
+
+ext.applyNativeModulesAppBuildGradle = { Project project, String packageName, String root = null ->
```
ArazAbishov: Upstream the change to the autolinking plugin to specify a package name in parameters. This way we can force the plugin to generate code for a module other than the one that `react-native config` was executed for. An alternative solution is to derive the `packageName` from the `project` that the plugin gets applied to.

android/app/src/main/java/com/sample/ComponentActivity.kt
```diff
@@ -0,0 +1,116 @@
...
+            "index.ios",
+            "main.ios",
```
tido64: `.android`
```suggestion
            "index.android",
            "main.android",
```

example/package.json
```diff
@@ -3,7 +3,8 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
-    "build:ios": "mkdirp dist && react-native bundle --entry-file index.js --platform ios --dev true --bundle-output dist/main.jsbundle --assets-dest dist --reset-cache"
+    "build:ios": "mkdirp dist && react-native bundle --entry-file index.js --platform ios --dev true --bundle-output dist/main.jsbundle --assets-dest dist --reset-cache", 
+    "build:android": "mkdirp dist && mkdirp dist/res && react-native bundle --entry-file index.js --platform android --dev true --bundle-output dist/main.jsbundle --assets-dest dist/res --reset-cache"
```
tido64: Nit: Sort
```suggestion
    "build:android": "mkdirp dist && mkdirp dist/res && react-native bundle --entry-file index.js --platform android --dev true --bundle-output dist/main.jsbundle --assets-dest dist/res --reset-cache",
    "build:ios": "mkdirp dist && react-native bundle --entry-file index.js --platform ios --dev true --bundle-output dist/main.jsbundle --assets-dest dist --reset-cache"
```

test-app-native-modules.gradle
```diff
@@ -0,0 +1,271 @@
+import groovy.json.JsonSlurper
```
tido64: Should we move all utility Gradle files into the `android` folder? In general, I think we should only leave the main one in root.

test-app-util.gradle
```diff
@@ -21,8 +21,8 @@ import java.nio.file.Paths
 * The search begins at the given base directory (a File object). The returned
 * path is a string.
 */
-ext.findNodeModulePath = { packageName ->
-    def basePath = rootDir.toPath().normalize()
+ext.findNodeModulePath = { baseDir, packageName ->
```
tido64: `findNodeModulesPath`?
```suggestion
ext.findNodeModulesPath = { baseDir, packageName ->
```

android/app/src/main/java/com/sample/ComponentActivity.kt
ArazAbishov: Fallback to the one provided by default in the `ReactNativeHost`

ArazAbishov:
Check whether hardcoded path delimiters in gradle scripts are compatible with windows. Otherwise use `File.separator`.

example/gradle.properties
ArazAbishov: https://github.com/microsoft/react-native-test-app/projects/1#card-35147280

test-app.gradle
ArazAbishov: https://github.com/microsoft/react-native-test-app/projects/1#card-35147280

test-app-native-modules.gradle
ArazAbishov: https://github.com/microsoft/react-native-test-app/projects/1#card-35147467

ArazAbishov:
> Check whether hardcoded path delimiters in gradle scripts are compatible with windows. Otherwise use `File.separator`.

Checked and it works fine.

android/app/src/main/java/com/sample/ComponentActivity.kt
```diff
@@ -0,0 +1,29 @@
+package com.sample
```
tido64: Do we have to use a proper namespace here?

android/app/src/main/java/com/sample/react/ReactBundleNameProvider.kt
```diff
@@ -0,0 +1,24 @@
+package com.sample.react
+
+import android.content.Context
+import javax.inject.Inject
+import javax.inject.Singleton
+
+private val possibleEntryFiles = listOf(
```
tido64: I'd move this into the function. It's not used elsewhere.

android/test-app-native-modules.gradle
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    /**
+     * Adds the react native modules as dependencies to the users `app` project
+     */
+    void addReactNativeModuleDependencies(Project appProject) {
+        reactNativeModules.forEach { reactNativeModule ->
+            def nameCleansed = reactNativeModule["nameCleansed"]
+            appProject.dependencies {
+                // TODO(salakar): are other dependency scope methods such as `api` required?
```
tido64: Was this copied from somewhere? What was its license? Do we need to put that copyright notice here?

android/test-app-native-modules.gradle
```diff
@@ -0,0 +1,282 @@
...
+        this.jsAppDir = jsAppDir
+        this.reactNativeModules = this.getReactNativeConfig()
+    }
+
+    /**
+     * Include the react native modules android projects and specify their project directory
+     */
+    void addReactNativeModuleProjects(DefaultSettings defaultSettings) {
+        reactNativeModules.forEach { reactNativeModule ->
+
```
tido64: Trim newline

example/package.json
```diff
@@ -3,7 +3,8 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
-    "build:ios": "mkdirp dist && react-native bundle --entry-file index.js --platform ios --dev true --bundle-output dist/main.jsbundle --assets-dest dist --reset-cache"
+    "build:android": "mkdirp dist && mkdirp dist/res && react-native bundle --entry-file index.js --platform android --dev true --bundle-output dist/main.jsbundle --assets-dest dist/res --reset-cache",
```
tido64:
```suggestion
    "build:android": "mkdirp dist/res && react-native bundle --entry-file index.js --platform android --dev true --bundle-output dist/main.jsbundle --assets-dest dist/res --reset-cache",
```

tido64: Why do the assets need to be inside `dist/res`?

.github/workflows/build.yml
```diff
@@ -51,3 +51,48 @@ jobs:
...
+        node-version: '12.9.1'  
+    - name: Install
+      run: |
+        yarn
+      working-directory: example
+    - name: Build
+      shell: cmd
+      run: yarn build:android                
+      working-directory: example
+    - name: Build android
```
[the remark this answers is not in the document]
ArazAbishov: Github Actions, or CMD, ignores the second line of the script. That's the reason why ended-up having two separate blocks.

android/app/build.gradle
```diff
@@ -19,20 +48,43 @@ android {
...
+    packagingOptions {
+        pickFirst "lib/armeabi-v7a/libc++_shared.so"
+        pickFirst "lib/arm64-v8a/libc++_shared.so"
+        pickFirst "lib/x86_64/libc++_shared.so"
+        pickFirst "lib/x86/libc++_shared.so"
+    }
 }
 
-def hermesEnginePath = findNodeModulePath("hermes-engine")
+// TODO: switch back to using path below when running on react-native v0.61.5
```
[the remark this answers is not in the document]
ArazAbishov: Yes. Do we want to be backwards compatible with older versions of RN? If not, I would rather not have additional logic for picking a package name.

android/app/src/main/java/com/sample/ComponentActivity.kt
ArazAbishov: Hm, what do you mean? The package name corresponds to the file system path, which is `com/sample` in this case :P

.github/workflows/build.yml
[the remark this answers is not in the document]
tido64: Why not run Bash then?

android/app/src/main/java/com/sample/react/ReactBundleNameProvider.kt
ArazAbishov: The reason to move it out is to avoid allocating the list every time the function is called (internally this list will be "java static"). The variable is also private to the file, so I would rather keep it as it is.

android/test-app-native-modules.gradle
ArazAbishov: That's correct. The whole file is a fork of the auto-linking plugin from the cli tools. I will add the header.

example/package.json
ArazAbishov: The assets option is ambiguous. In fact, it does not correspond to android assets, but rather to resources that should go under `res` folder.

android/app/src/main/java/com/sample/react/ReactBundleNameProvider.kt
tido64: Correct me if I'm wrong but doesn `lazy` imply that it only gets calculated once? I'd rather not have a list in static memory if possible.

android/app/src/main/java/com/sample/ComponentActivity.kt
tido64: I guess what I'm trying to ask is whether we should consider having a proper namespace. I know that it reflects the file structure.

ArazAbishov: Aha, I see. I prefer to namespace classes by "features" and when there is more than one class to put into a package. `ComponentActivity` does not have any logically correlated classes at the moment, but there will be a few in the future (when support for the native controllers lands). I would prefer to do it then, unless you have some other ideas?

android/app/src/main/java/com/sample/react/ReactBundleNameProvider.kt
ArazAbishov: That's correct, `lazy` in combination with the `ReactBundleNameProvider` being a singleton guarantee that the method will be executed once (unless you explicitly re-init the dependency graph).

The list itself, as well as its contents are immutable. It is safe to have it as static afaik. I will move it down to the method definition, but the difference here mostly comes down to the personal preference :P

.github/workflows/build.yml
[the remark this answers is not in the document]
ArazAbishov: Oh wow, it actually worked. Should have done that in the first place.. I will try to rework two definitions into one, since both of them follow the same steps.

android/app/build.gradle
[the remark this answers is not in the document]
tido64: If it adds extra complexity then let's wait till somebody raises an issue.

.github/workflows/build.yml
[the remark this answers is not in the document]
ArazAbishov: Holy crap, the single job definition worked too :D

.github/workflows/build.yml
```diff
@@ -51,3 +51,51 @@ jobs:
...
+      run: |
+        yarn
+      working-directory: example
+    - name: Build
+      run: |
+        set -eo pipefail
+        yarn build:android
+        sh gradlew clean build check test
+      working-directory: example
+  Android-windows:
```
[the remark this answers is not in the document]
ArazAbishov: It is not relevant anymore, as this build definition is gone :P

.github/workflows/build.yml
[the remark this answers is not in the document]
tido64: Yeah, this looks really nice now.

android/app/src/main/java/com/sample/ComponentActivity.kt
tido64: https://github.com/microsoft/react-native-test-app/issues/34

PR #25, opened 2020-03-20
Build Plop template on CI

tido64:


scripts/install-test-template.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -eo pipefail
+
+yarn
+yarn plop --dest template-example TemplateExample ${1:-All}
```
sweggersen: Should this be "all" (lower case)?

tido64: Good catch. I made it required.

PR #26, opened 2020-03-20
iOS: Prep for React Native 0.61

tido64:


test_app.rb
```diff
@@ -38,14 +38,31 @@ def resources_pod(package_root, current_dir = package_root)
   podspec_path = File.join(current_dir, 'ReactTestApp-Resources.podspec.json')
   File.write(podspec_path, spec.to_json())
   at_exit { File.delete(podspec_path) }
-  Pathname.new(current_dir).relative_path_from(package_root).to_s()
+  Pathname.new(current_dir).relative_path_from(project_root).to_s()
```
alloy: Just an FYI, but idiomatic Ruby is to omit parenthesis, when not passing arguments at least:

```suggestion
  Pathname.new(current_dir).relative_path_from(project_root).to_s
```

test_app.rb
```diff
@@ -38,14 +38,31 @@ def resources_pod(package_root, current_dir = package_root)
...
 end
 
-def use_test_app!(package_root)
+def use_react_native!(project_root)
+  react_native = Pathname.new(resolve_module('react-native'))
+
+  package_json = JSON.parse(File.read(File.join(react_native.to_s(), 'package.json')))
+  version = package_json['version'].match(/(?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)/)
+
+  if version[:major] == "0" and version[:minor] == "60"
```
alloy: FYI Most people avoid `and`, etc, as precedence can be unexpected. (I don’t recall the details, but must be googleable.)

test_app.rb
tido64: Thanks, I really appreciate these types of comments. Do you know of a good linter we can use?

alloy: 👍 Rubocop is the linter of choice.

PR #28, opened 2020-03-22
Run RuboCop on CI

tido64:


tido64:
> Didn’t you have this as part of the Gemfile? Because I think you want to lock this so you don’t suddenly get a different version and rules on CI.

Good point!

.github/workflows/build.yml
```diff
@@ -12,6 +12,8 @@ jobs:
       uses: actions/checkout@v2
     - name: minitest
       run: |
+        bundle install
+        rubocop
```
alloy: This should probably be prefixed with `bundle exec`, no? Same actually goes for the command below.

tido64: Thanks for pointing that out! I'm a total Ruby noob, so this is really helpful.

alloy: 🤗

PR #30, opened 2020-03-24
Test resources_pod()

tido64:


test/test_test_app.rb
```diff
@@ -0,0 +1,50 @@
...
+    ].each do |project_root|
+      podspec_path = resources_pod(project_root)
+      manifest_path = app_manifest_path(project_root, podspec_path)
+      manifest = JSON.parse(File.read(manifest_path))
+
+      assert_equal(2, manifest['resources'].length)
+      assert_equal('dist/assets', manifest['resources'][0])
+      assert_equal('dist/main.jsbundle', manifest['resources'][1])
+
+      File.delete(manifest_path)
```
alloy: iirc these assertions will raise an exception and when that happens this line is never reached. Doing this from a ‘finally’ clause will ensure it always happens.

```suggestion
  ensure
      File.delete(manifest_path)
```

test/test_test_app.rb
```diff
@@ -0,0 +1,50 @@
...
+      File.join(__dir__, 'fixtures', 'with_resources'),
+      File.join(__dir__, 'fixtures', 'with_resources', 'ios')
+    ].each do |project_root|
+      podspec_path = resources_pod(project_root)
+      manifest_path = app_manifest_path(project_root, podspec_path)
+      manifest = JSON.parse(File.read(manifest_path))
+
+      assert_equal(2, manifest['resources'].length)
+      assert_equal('dist/assets', manifest['resources'][0])
+      assert_equal('dist/main.jsbundle', manifest['resources'][1])
```
alloy: Just my 2c, so feel free to ignore; but I always prefer the following because then when the first count test fails you actually get to see the unexpected changes:

```ruby
assert_equal(%w{ dist/assets dist/main.jsbundle }, manifest['resources'].sort)
```

test/test_test_app.rb
tido64: Is this a new feature? RuboCop is yelling at me:
```
test/test_test_app.rb:47:3: E: Lint/Syntax: unexpected token kENSURE
(Using Ruby 2.3 parser; configure using TargetRubyVersion parameter, under AllCops)
  ensure
  ^^^^^^
```

alloy: Hmm, perhaps the ability to have each `do … end` block be implicit ‘try’ blocks is, yeah. Then do:

```ruby
begin
  # ...
ensure
  File.delete(manifest_path)
end
```

tido64: Thanks, that was it. The implicit behaviour was apparently introduced in 2.5.

alloy: Gotcha. It was always there in the scope of a method, which made the experience inconsistent, so glad they added it 👍

PR #40, opened 2020-03-26
Loading resources from the manifest on android

ArazAbishov:
* (wip) Forking resources between platforms in the manifest definition
* Loading and copying resources from the manifest on Android

### DIFF
--- a/example/Podfile.lock
+++ b/example/Podfile.lock
@@ -1,7 +1,7 @@
 PODS:
   - boost-for-react-native (1.63.0)
   - DoubleConversion (1.1.6)
-  - Example-Tests (0.0.5):
+  - Example-Tests (0.0.6):
     - React
   - Folly (2018.10.22.00):
     - boost-for-react-native
@@ -82,7 +82,7 @@ PODS:
     - React-Core (= 0.60.6)
   - React-RCTWebSocket (0.60.6):
     - React-Core (= 0.60.6)
-  - ReactTestApp-DevSupport (0.0.5)
+  - ReactTestApp-DevSupport (0.0.6)
   - ReactTestApp-Resources (1.0.0-dev)
   - SwiftLint (0.39.1)
   - yoga (0.60.6.React)
@@ -110,7 +110,7 @@ DEPENDENCIES:
   - React-RCTText (from `node_modules/react-native/Libraries/Text`)
   - React-RCTVibration (from `node_modules/react-native/Libraries/Vibration`)
   - React-RCTWebSocket (from `node_modules/react-native/Libraries/WebSocket`)
-  - ReactTestApp-DevSupport (from `./..`)
+  - ReactTestApp-DevSupport (from `./node_modules/react-native-test-app`)
   - ReactTestApp-Resources (from `.`)
   - SwiftLint
   - yoga (from `node_modules/react-native/ReactCommon/yoga`)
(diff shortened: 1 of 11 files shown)


example/app.json
```diff
@@ -6,8 +6,14 @@
       "displayName": "App"
     }
   },
-  "resources": [
-    "dist/assets",
-    "dist/main.jsbundle"
-  ]
+  "resources": {
```
tido64: I think we should be able to support the flattened structure for platform specific packages. E.g. for Android only packges:
```json
"resources": {
  "bundle": "dist/main.jsbundle",
  "resources": "dist/res"
}
```

ArazAbishov: Is that a good practice though? I think it is fine just to have ios or android objects if you want to consume the test app only for one platform.

ArazAbishov: On top of that, it will reduce complexity on our side.

test-app.gradle
```diff
@@ -20,24 +21,34 @@ ext.applyTestAppSettings = { DefaultSettings defaultSettings ->
...
+        if (bundlePath != null) {
+            from "${rootDir}/${bundlePath}"
+            into generatedAssetsDir
+        }
     }
 
     task copyResources(type: Copy) {
-        def resourcesPath = project.getProperties()
-                .get("testApp.resources")
+        def resourcesPath = resources["resources"]
```
tido64: Instead of having specific keys for resources and bundles, can we instead infer them by looking at the folder structure? E.g. if the path is `drawable-*/**/*` and `raw/**/*`, we can put them under `res/`, and put everything else under `assets/`? That way, we can use an array like for iOS and allow adding arbitrary assets.

example/app.json
tido64: I think it's a nice-to-have. Not sure how it would add complexity though. If the `android` key exists, use that; otherwise use `resources`. I don't think that would be too hard to add.

I would also like to get rid of the keys here and go with a flat structure (see comment below).

ArazAbishov: Right, but iOS for example, expects an array or strings, while android expects an object. I am just not comfortable with the idea of encoding these implicit rules into the manifest, rather than having it a little bit more "boilerplaty" and consistent.

test-app.gradle
ArazAbishov: Yes, we can do that, but then we will have to handle every android resource directory individually.

example/Podfile.lock
```diff
@@ -110,7 +110,7 @@ DEPENDENCIES:
   - React-RCTText (from `node_modules/react-native/Libraries/Text`)
   - React-RCTVibration (from `node_modules/react-native/Libraries/Vibration`)
   - React-RCTWebSocket (from `node_modules/react-native/Libraries/WebSocket`)
-  - ReactTestApp-DevSupport (from `./..`)
+  - ReactTestApp-DevSupport (from `./node_modules/react-native-test-app`)
```
tido64: You should use `yarn link` as described in the README. That will revert this line and the line below.

test-app.gradle
```diff
@@ -21,29 +22,54 @@ ext.applyTestAppSettings = { DefaultSettings defaultSettings ->
...
+    } else {
+        throw new IllegalArgumentException("The \"resources\" property of the app.json has " +
+            "to be either a list of paths, or must contain a nested list with the \"android\" key")
+    }
+
+    def androidResourceFiles = androidResources
+        .collect { file("$rootDir/$it") }
+
+    def androidResDir = androidResourceFiles
+        .find { it.isDirectory() && it.path.endsWith("res") }
```
tido64: Should we check that the last component of the path is `res`? I.e.
```suggestion
        .find { it.isDirectory() && it.name == "res" }
```

ArazAbishov:
image

PR #41, opened 2020-03-26
iOS: Generate Xcode project in nearest node_modules folder

tido64:


tido64:
> At some point it may be easier to convert path strings to `Pathname` instances at the input and keep them that way until you need the string representation.

I'll try looking into it in a separate PR.

PR #44, opened 2020-03-29
Android: Don't hard-code `app.json` path

tido64:


Brewfile
```diff
@@ -0,0 +1,5 @@
+tap 'AdoptOpenJDK/openjdk'
+brew 'clang-format'
+brew 'node'
+brew 'yarn'
+cask 'adoptopenjdk11'
```
ArazAbishov: I don't remember of a specific example when android toolchain breaks on JDK's other than v8, but it does happen fairly often. Hence, I would rather depend on v8 to avoid issues. 

Also JDK is often installed on the system, for example using sdk man. Is it possible to make this dependency optional? Installing multiple JDKs on the system can mess-up path variables.

android/test-app-util.gradle
```diff
@@ -1,5 +1,18 @@
...
+    if (currentDir == null) {
+        return null
+    }
+
+    def file = Paths.get(currentDir.toString(), fileName).toFile()
+    if (file.exists()) {
+        return file
+    }
+
+    return findFile(fileName, currentDir.getParent())
```
ArazAbishov: Use the `while` loop instead, as `findNodeModules` does?

plopfile.js
```diff
@@ -165,7 +165,7 @@ module.exports = (plop) => {
           template: [
             "rootProject.name='example'",
             "",
-            'apply from: file("${rootDir}/node_modules/react-native-test-app/test-app.gradle")',
+            `apply from: file("${exclusive ? "" : "../"}node_modules/react-native-test-app/test-app.gradle")`,
```
ArazAbishov: I think it would be better to use `findNodeModules` here instead. Then you don't have to differentiate between exclusive and non-exclusive modes.

Brewfile
tido64: Don’t you need JRE installed to run sdkmanager in the first place? This is for machines that don’t have it installed to begin with. I have a MacBook at home where I recently did a clean install of Catalina. It didn’t have JRE installed, and this was the version that worked for me. Other versions crashed or threw exceptions.

plopfile.js
tido64: Isn't this file something the consumer would write? Do they have access to that function here?

Brewfile
tido64: Tried `adoptopenjdk8` and it works, so I'll downgrade.

plopfile.js
ArazAbishov: That's a good point. There is an accessible utility function in the `root/android/test-app-util.gradle` file. Another question if we want it to be a part of the public API of the library.. Let's keep it as it is for now.

Issue #36, opened 2020-03-25
Support for native controllers on Android


ArazAbishov:
https://github.com/microsoft/react-native-test-app/pull/24

PR #46, opened 2020-03-31
Support for native controllers on Android

ArazAbishov:
Addresses https://github.com/microsoft/react-native-test-app/issues/36

Test app now can run fragments. Component name has to be FQCN of the fragment for it to work.


android/app/src/main/java/com/sample/component/NativeComponentActivity.kt
```diff
@@ -0,0 +1,73 @@
...
+import androidx.fragment.app.Fragment
+import com.facebook.react.ReactNativeHost
+import com.facebook.soloader.SoLoader
+import dagger.android.AndroidInjection
+import javax.inject.Inject
+
+class NativeComponentActivity : AppCompatActivity() {
+
+    @Inject
+    lateinit var reactNativeHost: ReactNativeHost
```
ArazAbishov: Ensure that `reactNativeHost` is receiving correct lifecycle events from the activity.

ArazAbishov:
Hold-on with review, I discovered a way to simplify the whole thing a lot.

android/app/src/main/java/com/sample/component/ComponentActivity.kt
```diff
@@ -0,0 +1,79 @@
...
+            }
+        }
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        SoLoader.init(this, false)
+
+        val componentDisplayName = intent.extras?.getString(COMPONENT_DISPLAY_NAME, null)
+            ?: throw IllegalArgumentException("Component display name has to be provided.")
```
tido64: Can't we just fallback to `componentName`? I don't think it should be required.

android/app/src/main/java/com/sample/component/ComponentActivity.kt
```diff
@@ -0,0 +1,79 @@
...
+        val componentName = intent.extras?.getString(COMPONENT_NAME, null)
+            ?: throw IllegalArgumentException("Class name has to be provided.")
+
+        title = componentDisplayName
```
tido64: Nit: Do we really need the temporary variable?

android/app/src/main/java/com/sample/component/ComponentActivity.kt
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        SoLoader.init(this, false)
+
+        val componentDisplayName = intent.extras?.getString(COMPONENT_DISPLAY_NAME, null)
+            ?: throw IllegalArgumentException("Component display name has to be provided.")
+        val componentName = intent.extras?.getString(COMPONENT_NAME, null)
+            ?: throw IllegalArgumentException("Class name has to be provided.")
```
tido64: Can we move these closer to usage site?

android/app/src/main/java/com/sample/component/ComponentActivity.kt
```diff
@@ -0,0 +1,79 @@
...
+        if (android.R.id.home == item.itemId) {
+            onBackPressed()
+            return true;
+        }
+
+        return super.onOptionsItemSelected(item)
+    }
+
+    override fun getSystemService(name: String): Any? {
+        if (name == "service:reactNativeHostService") {
```
tido64: What does this do?

android/app/src/main/java/com/sample/component/ComponentActivity.kt
ArazAbishov: Is it? If `displayName` is optional, I will have to change the manifest types as well to make the field optional.

android/app/src/main/java/com/sample/component/ComponentActivity.kt
ArazAbishov: If the fragment that is about to be attached will need to access the `ReactNativeHost`, it will be able to pull it out through the context.

android/app/src/main/java/com/sample/component/ComponentActivity.kt
tido64: _One could also forgo the name, in which case the `entryPoint` string will be used instead._

-- https://github.com/tido64/discussions-and-proposals/blob/tido/test-app-as-package/proposals/0000-React-Native-Test-App-as-a-Package.md#the-manifest

tido64:
@ArazAbishov: RE: build failures, should we cache Gradle if the download often fails? https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows

PR #58, opened 2020-04-08
Clean up Example app folder

### LINKED ISSUE #33: Move platform specific files in `example` to `example/{android,ios}`
See https://github.com/microsoft/react-native-test-app/pull/24#pullrequestreview-381347263

tido64:
Resolves #33.


example/package.json
```diff
@@ -12,6 +12,9 @@
   },
   "devDependencies": {
     "@babel/core": "^7.0.0",
+    "@react-native-community/cli": "^3.0.0",
```
ArazAbishov, line 15: What was the reason to bump the version?

tido64: 2.x doesn't support this folder structure. It fails to find `react-native`.

PR #60, opened 2020-04-09
Add ESLint

tido64:


tido64:
Annotations work but is a bit cumbersome to find. I thought they would surface on the PR as a comment:
Screenshot 2020-04-09 at 16 54 13

PR #63, opened 2020-04-11
Android: Add menu for changing JS bundle source

### LINKED ISSUE #53: Android: options to load the bundle either from the dev server or assets
The iOS test-app has two UI options to choose the bundle source. Android should have support for the same feature. 

image

tido64:
Incidentally, this also fixes dev server support.

Resolves #53.

### DIFF
--- a/android/app/src/main/java/com/sample/MainActivity.kt
+++ b/android/app/src/main/java/com/sample/MainActivity.kt
@@ -42,19 +42,19 @@ class MainActivity : ReactActivity() {
 
         val manifest = manifestProvider.manifest
                 ?: throw IllegalStateException("app.json is not provided or TestApp is misconfigured")
-        findViewById<MaterialToolbar>(R.id.topAppBar).apply {
+        findViewById<MaterialToolbar>(R.id.top_app_bar).apply {
             title = manifest.displayName
             setOnMenuItemClickListener { menuItem ->
                 when (menuItem.itemId) {
-                    R.id.loadEmbeddedJsBundle -> {
+                    R.id.load_embedded_js_bundle -> {
                         reload(this, true)
                         true
                     }
-                    R.id.loadFromDevServer -> {
+                    R.id.load_from_dev_server -> {
                         reload(this, false)
                         true
                     }
-                    R.id.showDevOptions -> {
+                    R.id.show_dev_options -> {
                         reactInstanceManager.devSupportManager.showDevOptionsDialog()
                         true
                     }
--- a/android/app/src/main/java/com/sample/MainActivity.kt
+++ b/android/app/src/main/java/com/sample/MainActivity.kt
@@ -6,10 +6,13 @@ import androidx.recyclerview.widget.DividerItemDecoration
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
 import com.facebook.react.ReactActivity
+import com.facebook.soloader.SoLoader
+import com.google.android.material.appbar.MaterialToolbar
 import com.sample.component.ComponentActivity
 import com.sample.component.ComponentListAdapter
 import com.sample.component.ComponentViewModel
 import com.sample.manifest.ManifestProvider
+import com.sample.react.TestAppReactNativeHost
 import dagger.android.AndroidInjection
 import javax.inject.Inject
 
@@ -20,9 +23,9 @@ class MainActivity : ReactActivity() {
 
     private val listener = { component: ComponentViewModel ->
         startActivity(
-            ComponentActivity.newIntent(
-                this, component.name, component.displayName
-            )
+                ComponentActivity.newIntent(
+                        this, component.name, component.displayName
+                )
         )
     }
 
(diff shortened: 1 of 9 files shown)


tido64:
Screenshots:
image
image
image

android/app/src/main/java/com/sample/MainActivity.kt
```diff
@@ -32,21 +35,67 @@ class MainActivity : ReactActivity() {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_main)
 
+        val reactNativeHost = reactNativeHost
+        if (reactNativeHost is TestAppReactNativeHost) {
+            reactNativeHost.currentActivity = this
```
ArazAbishov, line 40: Setting `currentActivity` is taken care of by `ReactActivity` that is extended by `MainActivity`. Is there a particular reason to set here explicitly?

ArazAbishov: `ReactActivity.onResume -> ReactActivityDelegate.onResume -> ReactInstanceManager.onHostResume`:

```
  @ThreadConfined(UI)
  public void onHostResume(Activity activity) {
    UiThreadUtil.assertOnUiThread();

    mCurrentActivity = activity;
    ...
}
```

android/app/src/main/java/com/sample/MainActivity.kt
```diff
@@ -32,21 +35,67 @@ class MainActivity : ReactActivity() {
...
+        }
+
         val manifest = manifestProvider.manifest
-            ?: throw IllegalStateException("app.json is not provided or TestApp is misconfigured")
+                ?: throw IllegalStateException("app.json is not provided or TestApp is misconfigured")
+        findViewById<MaterialToolbar>(R.id.topAppBar).apply {
```
ArazAbishov, line 45: Nit: it would be great to extract this call into own `setupToolbar(displayName: String)` method to declutter the `onCreate()` a little bit. The same applies to the recyclerView setup. I can do that, just let me know if you're okay with it :)

android/app/src/main/java/com/sample/MainActivity.kt
```diff
@@ -32,21 +35,67 @@ class MainActivity : ReactActivity() {
...
             adapter = ComponentListAdapter(
-                LayoutInflater.from(context), components, listener
+                    LayoutInflater.from(context), components, listener
             )
 
             addItemDecoration(DividerItemDecoration(context, DividerItemDecoration.VERTICAL))
         }
     }
+
+    private fun reload(toolbar: MaterialToolbar, useEmbeddedBundle: Boolean) {
```
ArazAbishov, line 79: Nit: extract this method into `TestAppReactNativeHost`, except the piece responsible for the Toolbar? This way we can re-use from other parts of the app if needed + simplifies this activity.

android/app/src/main/java/com/sample/react/TestAppReactNativeHost.kt
```diff
@@ -1,22 +1,47 @@
...
 class TestAppReactNativeHost @Inject constructor(
-    application: Application,
-    private val reactBundleNameProvider: ReactBundleNameProvider
+        application: Application,
+        private val reactBundleNameProvider: ReactBundleNameProvider
 ) : ReactNativeHost(application) {
 
-    override fun getBundleAssetName() =
-        reactBundleNameProvider.bundleName ?: super.getBundleAssetName()
+    var currentActivity: ReactActivity? = null
```
ArazAbishov, line 17: As said earlier, I don't think we need to keep track of currentActivity `ourselves`.

android/app/src/main/res/layout/activity_main.xml
```diff
@@ -1,9 +1,30 @@
...
-    android:paddingBottom="8dp" />
+    android:layout_height="match_parent">
+
+    <com.google.android.material.appbar.AppBarLayout
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        app:liftOnScroll="true">
+
+        <com.google.android.material.appbar.MaterialToolbar
+            android:id="@+id/topAppBar"
```
ArazAbishov, line 13: Nit: please, use snake_case for ids.

android/app/src/main/res/menu/top_app_bar.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto">
+    <item
+        android:id="@+id/loadEmbeddedJsBundle"
```
ArazAbishov, line 5: Nite: snake_case here as well, and for items below.

android/app/src/main/res/layout/activity_main.xml
tido64: I'm really confused about this. The Material Design doc uses camelCase: https://material.io/develop/android/components/top-app-bars/. It's very inconsistent throughout. Is snake_case the one true way?

android/app/src/main/java/com/sample/MainActivity.kt
tido64: Yes, I get NPEs otherwise. I don't think `onHostResume` gets called on initialisation. Not on my device at least.

android/app/src/main/res/layout/activity_main.xml
ArazAbishov: Google is not very consistent in their examples. It really comes down to the personal preference, and I think that having naming of `ids` consistent with the rest of resources makes sense. For example, when you reference layouts, string resources, dimens, etc, all of them are defined in the snake case.

android/app/src/main/java/com/sample/MainActivity.kt
tido64: Let's do that in a separate PR.

android/app/src/main/java/com/sample/MainActivity.kt
ArazAbishov: It doesn't set the `currentActivity` in `ReactActivityDelegate.onResume` because it can't find an instance of `ReactInstanceManager`, that is lazily created whenever you call `reactNativeHost.reactInstanceManager`. The way how we can indirectly initialise it is by calling `reactInstanceManager`. 

I see this working pretty well with another PR of yours for setting runtime info. We can poll the JS executor prop out of `ReactInstanceManager` instead of defining a custom build prop :) 

```
findViewById<TextView>(R.id.runtimeInfo).text =
            reactNativeHost.reactInstanceManager?.jsExecutorName
```

This way we won't need to have `currentActivity`, and to override `createReactInstanceManager` :)

tido64: Right, I see. In my opinion, we either keep lazy instantiating `ReactInstanceManager` or we do it earlier. I don't want to rely on an implicit behaviour like getting the JS executor name. Let's address this in a follow-up PR?

ArazAbishov: Sure, we can explicitly initialise `ReactInstanceManager` by calling `createReactInstanceManager()` within `TestAppReactNativeHost` for example. Let's look at it in the separate PR.

PR #64, opened 2020-04-11
Android: Show runtime information

tido64:


tido64:
Screenshot:
image

android/app/src/main/java/com/sample/MainActivity.kt
```diff
@@ -74,6 +76,15 @@ class MainActivity : ReactActivity() {
 
             addItemDecoration(DividerItemDecoration(context, DividerItemDecoration.VERTICAL))
         }
+
+        findViewById<TextView>(R.id.runtimeInfo).apply {
+            text = resources.getString(
+                    R.string.runtime_info,
+                    (ReactNativeVersion.VERSION["major"] as Number).toInt(),
```
ArazAbishov, line 83: Assuming that `ReactNativeHost.VERSION["key"]` returns you a string, it is sufficient just to call `toInt()` directly on it.

ArazAbishov, line 83:
```suggestion
                    ReactNativeVersion.VERSION["major"] as Int,
```

android/app/src/main/java/com/sample/MainActivity.kt
```diff
@@ -74,6 +76,15 @@ class MainActivity : ReactActivity() {
...
+                    (ReactNativeVersion.VERSION["minor"] as Number).toInt(),
```
ArazAbishov, line 84:
```suggestion
                    ReactNativeVersion.VERSION["minor"] as Int,
```

android/app/src/main/java/com/sample/MainActivity.kt
```diff
@@ -74,6 +76,15 @@ class MainActivity : ReactActivity() {
...
+                    (ReactNativeVersion.VERSION["patch"] as Number).toInt(),
```
ArazAbishov, line 85:
```suggestion
                    ReactNativeVersion.VERSION["patch"] as Int,
```

android/app/src/main/res/layout/activity_main.xml
```diff
@@ -27,4 +27,12 @@
         android:paddingBottom="0dp"
         app:layout_behavior="@string/appbar_scrolling_view_behavior" />
 
+    <TextView
+        android:id="@+id/runtimeInfo"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_gravity="bottom"
+        android:gravity="end"
+        android:padding="4dp" />
```
ArazAbishov, line 36:
```suggestion
        android:layout_margin="16dp" />
```

android/app/src/main/java/com/sample/react/TestAppReactNativeHost.kt
```diff
@@ -15,21 +16,33 @@ class TestAppReactNativeHost @Inject constructor(
         private val reactBundleNameProvider: ReactBundleNameProvider
 ) : ReactNativeHost(application) {
 
-    var currentActivity: ReactActivity? = null
+    private var currentActivity: Activity? = null
```
tido64, line 19: @ArazAbishov: Looks like I can't get rid of this. The Back button won't behave correctly if we don't provide the correct activity on reload, e.g. it will close the app if you press back from a `ComponentActivity`.

android/app/src/main/java/com/sample/MainActivity.kt
```diff
@@ -36,7 +38,7 @@ class MainActivity : ReactActivity() {
 
         val reactNativeHost = reactNativeHost
         if (reactNativeHost is TestAppReactNativeHost) {
-            reactNativeHost.currentActivity = this
+            reactNativeHost.startInBackground()
```
ArazAbishov, line 41: It will crash after configuration change (rotate the phone, for example) because of the assertion. We should check if the instance is already started before calling `startInBackground()` here.

android/app/src/main/java/com/sample/react/TestAppReactNativeHost.kt
```diff
@@ -43,7 +56,11 @@ class TestAppReactNativeHost @Inject constructor(
                 .setJSMainModulePath(jsMainModuleName)
                 .addPackages(packages)
                 .setUseDeveloperSupport(useDeveloperSupport && !useEmbeddedBundle)
-                .setInitialLifecycleState(LifecycleState.RESUMED)
+                .setInitialLifecycleState(if (currentActivity == null) {
+                    LifecycleState.BEFORE_CREATE
```
ArazAbishov, line 60: This is ambiguous, as we don't know if activity going to be created at all if it is null. I can accept it as long as we're going to refactor this to rely on the `ReactActivityDelegate` to manage the `currentActivity` instance later.

tido64: Isn't that fine though? Why would RN care whether the Activity gets created if the initial state is correct?

ArazAbishov: I checked the source code of the `ReactNativeHost`, and they make the same assumption on the activity state. My concern was that you can set the state to `BEFORE_CREATE`, but then activity is actually never created and you end-up in some weird state.

android/app/src/main/java/com/sample/react/TestAppReactNativeHost.kt
ArazAbishov: Okay, just keep it here for now. Could you please create a ticket, so I can take a look at it later?

tido64: All yours: #66

PR #68, opened 2020-04-17
Avoid crashing when bundle is not packaged

ArazAbishov:
How to reproduce the issue on master: do not run `yarn build:android` and launch the example app. 

Fix: when bundle is not supplied, we have to enable the dev mode.

### DIFF
--- a/android/app/src/main/java/com/sample/react/TestAppReactNativeHost.kt
+++ b/android/app/src/main/java/com/sample/react/TestAppReactNativeHost.kt
@@ -2,7 +2,10 @@ package com.sample.react
 
 import android.app.Activity
 import android.app.Application
-import com.facebook.react.*
+import com.facebook.react.PackageList
+import com.facebook.react.ReactInstanceManager
+import com.facebook.react.ReactNativeHost
+import com.facebook.react.ReactPackage
 import com.facebook.react.bridge.ReactMarker
 import com.facebook.react.bridge.ReactMarkerConstants
 import com.facebook.react.common.LifecycleState
@@ -12,8 +15,8 @@ import javax.inject.Singleton
 
 @Singleton
 class TestAppReactNativeHost @Inject constructor(
-        application: Application,
-        private val reactBundleNameProvider: ReactBundleNameProvider
+    application: Application,
+    private val reactBundleNameProvider: ReactBundleNameProvider
 ) : ReactNativeHost(application) {
 
     private var currentActivity: Activity? = null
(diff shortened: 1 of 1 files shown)


android/app/src/main/java/com/sample/react/TestAppReactNativeHost.kt
```diff
@@ -12,8 +15,8 @@ import javax.inject.Singleton
 
 @Singleton
 class TestAppReactNativeHost @Inject constructor(
-        application: Application,
```
tido64, line 15: Android Studio added the extra indentation. Do you have different setting or something?

android/app/src/main/java/com/sample/react/TestAppReactNativeHost.kt
```diff
@@ -49,32 +52,34 @@ class TestAppReactNativeHost @Inject constructor(
     override fun createReactInstanceManager(): ReactInstanceManager {
         ReactMarker.logMarker(ReactMarkerConstants.BUILD_REACT_INSTANCE_MANAGER_START)
         val reactInstanceManager = ReactInstanceManager.builder()
-                .setApplication(application)
-                .setCurrentActivity(currentActivity)
-                .setJavaScriptExecutorFactory(javaScriptExecutorFactory)
-                .setBundleAssetName(bundleAssetName)
-                .setJSMainModulePath(jsMainModuleName)
-                .addPackages(packages)
-                .setUseDeveloperSupport(useDeveloperSupport && !useEmbeddedBundle)
```
tido64, line 58: This is going to break switching between embedded and dev server.

tido64: Maybe we can determine the initial value of `useEmbeddedValue` based on `bundleAssetName != null` instead?

android/app/src/main/java/com/sample/react/TestAppReactNativeHost.kt
ArazAbishov: I noticed the difference too, and double checked the code style. Mine is set to the Kotlin Guidelines code style and it formats files as you see in this PR. Can you check what happens on your side?

android/app/src/main/java/com/sample/react/TestAppReactNativeHost.kt
ArazAbishov: Right, I reproduced the crash. Initialising to `bundleAssetName != null` sounds like an idea.

android/app/src/main/java/com/sample/react/TestAppReactNativeHost.kt
tido64: Continuation indent is set to 8. I haven't changed this so I'm assuming it's default.
image

ArazAbishov: Can you click on `Set From -> Predefined style -> Kotlin style guide`?

ArazAbishov:
Abandoning this one in favor of another PR that also includes the fix.

android/app/src/main/java/com/sample/react/TestAppReactNativeHost.kt
tido64: No diff

ArazAbishov: Crap.. Which version of AS?

tido64: 3.6.2

ArazAbishov: I am running on 3.6.1, let me updated and check the new version.

ArazAbishov: Updated, no diff. Reset to defaults, and then imported the kotlin style, no diff :D

tido64: Can you close AS, remove `~/Library/Preferences/AndroidStudio3.6`, then try again?

ArazAbishov: Well, that bricked my AS installation :D

ArazAbishov: After re-installing the AS, **default** code style was changing formatting similar to what you had, but if you set the kotlin style, then it will leave it as it is.

ArazAbishov: You should try first to reset the style to default, and then apply the kotlin style guide.

tido64: Can we please use whatever comes out of box? I don't want all contributors to have to do this.

ArazAbishov: Well, this *is*, in fact, is the android style guide: https://developer.android.com/kotlin/style-guide

ArazAbishov: For some reason, it is not set by default in AS, even though they ship the style.

tido64: Copy/pasting the comment from the other PR: Can we check in the settings so we don't have to ask contributors to change their settings in AS?

ArazAbishov: Ah, wait, crap, they've switched it around sometime ago and I haven't noticed. Now the style guide in the settings is this one: https://kotlinlang.org/docs/reference/coding-conventions.html. Oops..

Issue #66, opened 2020-04-14
Investigate whether we can remove `TestAppReactNativeHost.currentActivity`


tido64:
See comment in https://github.com/microsoft/react-native-test-app/pull/64#discussion_r408134563.

PR #69, opened 2020-04-17
Refactor the TestAppReactNativeHost

ArazAbishov:
Addresses https://github.com/microsoft/react-native-test-app/issues/66 and fixes the crash when bundle is not present. Also proactively disables the menu item when bundle is not available.

### DIFF
--- a/android/app/src/main/java/com/sample/MainActivity.kt
+++ b/android/app/src/main/java/com/sample/MainActivity.kt
@@ -28,8 +28,8 @@ class MainActivity : ReactActivity() {
     @Inject
     lateinit var bundleNameProvider: ReactBundleNameProvider
 
-    @Inject
-    lateinit var testReactNativeVersion: TestAppReactNativeHost
+    private val testReactNativeVersion: TestAppReactNativeHost
+        get() = reactNativeHost as TestAppReactNativeHost
 
     private val listener = { component: ComponentViewModel ->
         startActivity(
--- a/android/app/src/main/java/com/sample/MainActivity.kt
+++ b/android/app/src/main/java/com/sample/MainActivity.kt
@@ -33,9 +33,9 @@ class MainActivity : ReactActivity() {
 
     private val listener = { component: ComponentViewModel ->
         startActivity(
-            ComponentActivity.newIntent(
-                this, component.name, component.displayName
-            )
+                ComponentActivity.newIntent(
+                        this, component.name, component.displayName
+                )
         )
     }
 
@@ -46,7 +46,7 @@ class MainActivity : ReactActivity() {
         setContentView(R.layout.activity_main)
 
         val manifest = manifestProvider.manifest
-            ?: throw IllegalStateException("app.json is not provided or TestApp is misconfigured")
+                ?: throw IllegalStateException("app.json is not provided or TestApp is misconfigured")
 
         setupToolbar(manifest.displayName)
         setupRecyclerView(manifest.components)
@@ -96,7 +96,7 @@ class MainActivity : ReactActivity() {
         findViewById<RecyclerView>(R.id.recyclerview).apply {
             layoutManager = LinearLayoutManager(context)
             adapter = ComponentListAdapter(
-                LayoutInflater.from(context), components, listener
+                    LayoutInflater.from(context), components, listener
             )
 
             addItemDecoration(DividerItemDecoration(context, DividerItemDecoration.VERTICAL))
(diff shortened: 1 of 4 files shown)


android/app/src/main/java/com/sample/react/TestAppReactNativeHost.kt
```diff
@@ -2,45 +2,76 @@ package com.sample.react
...
+    enum class Action {
+        RESTART, RELOAD, NOOP
+    }
+
+    abstract fun moveTo(to: BundleSource): Action
+
+    object Disk : BundleSource() {
+        override fun moveTo(to: BundleSource): Action {
+            return when (to) {
+                Disk -> Action.NOOP
```
tido64, line 26: This used to restart RN. I think that's still a useful feature to keep.

android/app/src/main/java/com/sample/MainActivity.kt
```diff
@@ -22,11 +25,17 @@ class MainActivity : ReactActivity() {
     @Inject
     lateinit var manifestProvider: ManifestProvider
 
+    @Inject
+    lateinit var bundleNameProvider: ReactBundleNameProvider
+
+    @Inject
+    lateinit var testReactNativeVersion: TestAppReactNativeHost
```
tido64, line 32: `ReactActivity` has a getter for `ReactNativeHost`. Can't we just add a getter that casts it to the correct type?

android/app/src/main/java/com/sample/MainActivity.kt
```diff
@@ -22,11 +25,17 @@ class MainActivity : ReactActivity() {
...
+    private val testAppReactNativeHost: TestAppReactNativeHost
+        get() = reactNativeHost as TestAppReactNativeHost
+
     private val listener = { component: ComponentViewModel ->
         startActivity(
-                ComponentActivity.newIntent(
-                        this, component.name, component.displayName
-                )
+            ComponentActivity.newIntent(
+                this, component.name, component.displayName
```
tido64, line 37: Can we settle on the default AS settings (whatever they are) to avoid unnecessary whitespace changes in the future?

android/app/src/main/java/com/sample/react/TestAppReactNativeHost.kt
```diff
@@ -2,45 +2,69 @@ package com.sample.react
...
 
+sealed class BundleSource {
+    enum class Action {
+        RESTART, RELOAD
+    }
+
+    abstract fun moveTo(to: BundleSource): Action
+
+    object Disk : BundleSource() {
+        override fun moveTo(to: BundleSource) = Action.RESTART
```
tido64, line 24: IIRC, it should be sufficient to do `Action.RELOAD` for Disk -> Server.

android/app/src/main/java/com/sample/react/TestAppReactNativeHost.kt
```diff
@@ -2,45 +2,69 @@ package com.sample.react
...
         }
 
-        clear()
+        val action = source.moveTo(newSource)
+        source = newSource
 
-        this.currentActivity = activity
-        this.useEmbeddedBundle = useEmbeddedBundle
+        when (action) {
+            BundleSource.Action.RELOAD -> reactInstanceManager.devSupportManager.handleReloadJS()
```
tido64, line 54: Nit: Can we add braces to make the block consistent?
```suggestion
            BundleSource.Action.RELOAD -> {
                reactInstanceManager.devSupportManager.handleReloadJS()
            }
```

android/app/src/main/java/com/sample/MainActivity.kt
ArazAbishov: We should stick to the kotlin style guide, and I believe that the changes here are following that style. See my other comment in the closed PR.

android/app/src/main/java/com/sample/react/TestAppReactNativeHost.kt
ArazAbishov: I believe that when you switch to loading from the server, you have to rebuild the `ReactInstanceManager` with the correct `useDeveloperSupport` flag set and the asset name.

ArazAbishov: In other words, I would leave it as it is.

android/app/src/main/java/com/sample/react/TestAppReactNativeHost.kt
ArazAbishov: Done!

android/app/src/main/java/com/sample/MainActivity.kt
tido64: Can we check in the settings so we don't have to ask contributors to change their settings in AS?

tido64: (I'm fine with a new PR for that.)

ArazAbishov: Reverted the style changes.

PR #70, opened 2020-04-17
Add EditorConfig and format all files

tido64:


android/app/src/main/java/com/sample/MainActivity.kt
```diff
@@ -96,19 +96,19 @@ class MainActivity : ReactActivity() {
...
         }
 
         findViewById<TextView>(R.id.runtime_info).apply {
             text = resources.getString(
-                    R.string.runtime_info,
-                    ReactNativeVersion.VERSION["major"] as Int,
-                    ReactNativeVersion.VERSION["minor"] as Int,
-                    ReactNativeVersion.VERSION["patch"] as Int,
-                    reactInstanceManager.jsExecutorName
+                R.string.runtime_info,
```
ArazAbishov, line 107: It seems that the continuation indent setting is 4 spaces instead of 8 compared to the default IDE settings. Could you please set the `continuation_indent_size=8` and reformat files?

PR #74, opened 2020-04-23
iOS: Use path of `Podfile` as project root

tido64:


test_app.rb
```diff
@@ -22,6 +22,15 @@ def find_file(file_name, current_dir)
   find_file(file_name, current_dir.parent)
 end
 
+def find_project_root
+  podfile_path = Thread.current.backtrace_locations.find do |location|
```
alloy, line 26: Oh nice one! TIL. I always used `caller`, but that would require parsing the paths from the frames.

tido64: I also found [`caller_locations`](https://ruby-doc.org/core-2.7.1/Kernel.html#method-i-caller_locations) earlier today but I had already submitted the PR ¯\\_(ツ)_/¯

Anyway, thanks for the review ❤️

alloy: Even that one is news to me 😄 It’s been too long since I’ve done a good amount of Ruby with the latest versions.

PR #76, opened 2020-04-24
ci(github): Set up release pipeline

tido64:


tido64:
How does it work: https://github.com/semantic-release/semantic-release#how-does-it-work

PR #83, opened 2020-05-15
fix: inconsistent component order on home screen

tido64:
Implements https://github.com/microsoft/continuity/issues/61#issuecomment-625212973.

### DIFF
--- a/android/app/src/main/java/com/sample/manifest/Manifest.kt
+++ b/android/app/src/main/java/com/sample/manifest/Manifest.kt
@@ -2,14 +2,16 @@ package com.sample.manifest
 
 import com.squareup.moshi.JsonClass
 
-typealias Components = Map<String, Component>
-
 @JsonClass(generateAdapter = true)
 data class Manifest(
     val name: String,
     val displayName: String,
-    val components: Components
+    val components: List<Component>
 )
 
 @JsonClass(generateAdapter = true)
-data class Component(val displayName: String?)
+data class Component(
+    val appKey: String,
+    val displayName: String?,
+    val initialProperties: Map<String, String?>?
+)
--- a/ios/ReactTestApp/ContentView.swift
+++ b/ios/ReactTestApp/ContentView.swift
@@ -49,8 +49,8 @@ final class ContentViewController: UITableViewController {
 
         DispatchQueue.global(qos: .userInitiated).async { [weak self] in
             self?.reactInstance.initReact { components in
-                let items: [NavigationLink] = components.map { (name, component) in
-                    (component.displayName ?? name, { self?.navigate(to: component, name: name) })
+                let items: [NavigationLink] = components.map { (component) in
+                    (component.displayName ?? component.appKey, { self?.navigate(to: component) })
                 }
                 DispatchQueue.main.async {
                     guard let strongSelf = self else {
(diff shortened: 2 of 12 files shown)


android/app/src/main/java/com/sample/manifest/Manifest.kt
```diff
@@ -2,14 +2,20 @@ package com.sample.manifest
 
 import com.squareup.moshi.JsonClass
 
-typealias Components = Map<String, Component>
+typealias Components = Array<Component>
```
ArazAbishov, line 5: The benefits of having a type alias are not that significant with removal of the `Map`. Thus, I recommend to remove it. Also, it is not common to use `Array` as it doesn't implement an immutable collection API. I would replace with `List<T>` instead.

android/app/src/main/java/com/sample/manifest/Manifest.kt
```diff
@@ -2,14 +2,20 @@ package com.sample.manifest
...
 
 @JsonClass(generateAdapter = true)
 data class Manifest(
     val name: String,
     val displayName: String,
+
+    @Suppress("ArrayInDataClass")
```
ArazAbishov, line 12: I suspect that this warning will be gone if `Array` is replaced with `List<T>` as suggested in the comment above.

android/app/src/main/java/com/sample/manifest/Manifest.kt
```diff
@@ -2,14 +2,20 @@ package com.sample.manifest
...
     val components: Components
 )
 
 @JsonClass(generateAdapter = true)
-data class Component(val displayName: String?)
+data class Component(
+    val appKey: String,
+    val displayName: String?,
+    val initialProperties: Map<String, String>?
```
ArazAbishov, line 20: Can the prop value be `null`? If yes, the type of `initialProperties` needs to reflect that: `Map<String, String?>?`

ios/ReactTestApp/ContentView.swift
```diff
@@ -49,8 +49,8 @@ final class ContentViewController: UITableViewController {
 
         DispatchQueue.global(qos: .userInitiated).async { [weak self] in
             self?.reactInstance.initReact { components in
-                let items: [NavigationLink] = components.map { (name, component) in
-                    (component.displayName ?? name, { self?.navigate(to: component, name: name) })
+                let items: [NavigationLink] = components.map { (component) in
+                    (component.displayName ?? component.appKey, { self?.navigate(to: component) })
```
ArazAbishov, line 53: I am not sure if I understand this expression correctly, so feel free to dismiss this comment if I am wrong. Shouldn't the `component.appKey` null-checked first and then `displayName`?

ios/ReactTestApp/ContentView.swift
```diff
@@ -132,21 +132,21 @@ final class ContentViewController: UITableViewController {
...
-    private func navigate(to component: Component, name: String) {
+    private func navigate(to component: Component) {
         guard let bridge = reactInstance.bridge,
               let navigationController = navigationController else {
             return
         }
 
         let viewController: UIViewController = {
-            if let viewController = RTAViewControllerFromString(name, bridge) {
+            if let viewController = RTAViewControllerFromString(component.appKey, bridge) {
```
ArazAbishov, line 142: Shouldn't we fallback to `displayName` here too if we want to keep backwards compat?

ios/ReactTestApp/ContentView.swift
tido64: `appKey` is what used to be the keys under `components`. It is required.

ios/ReactTestApp/ContentView.swift
tido64: No, this function tries to retrieve the native view controller given its name.

PR #84, opened 2020-05-15
fix(android): Forward initial properties from manifest

tido64:


android/app/src/main/java/com/sample/component/ComponentActivity.kt
```diff
@@ -6,25 +6,53 @@ import android.os.Bundle
 import android.view.MenuItem
 import androidx.fragment.app.Fragment
 import com.facebook.react.ReactActivity
+import com.facebook.react.ReactActivityDelegate
 
 class ComponentActivity : ReactActivity() {
 
+    class ComponentActivityDelegate(
```
ArazAbishov, line 13: You can change this class to be `inner` then you won't need to pass `initialProperties` as a constructor property. The class should also be private:`private inner ComponentActivity: ReactActivity()`

android/app/src/main/java/com/sample/component/ComponentActivity.kt
```diff
@@ -6,25 +6,53 @@ import android.os.Bundle
...
+        activity: ReactActivity,
+        mainComponentName: String?,
+        private val initialProperties: Bundle?
```
ArazAbishov, line 16: As I said in the comment above, you won't need this prop.

android/app/src/main/java/com/sample/component/ComponentActivity.kt
ArazAbishov: Scrap that. I think you can just avoid having a prop inside of `ComponentActivity` altogether by:

```kotlin
private inner class ComponentActivityDelegate(
    activity: ReactActivity,
    mainComponentName: String?,
) : ReactActivityDelegate(activity, mainComponentName) {
    override fun getLaunchOptions(): Bundle? {
      return intent.extras?.getBundle(COMPONENT_INITIAL_PROPERTIES)
    }
}
```

android/app/src/main/java/com/sample/component/ComponentActivity.kt
```diff
@@ -6,25 +6,53 @@ import android.os.Bundle
...
+                    for ((k, v) in componentInitialProperties) {
+                        bundle.putString(k, v)
+                    }
+                    putExtra(COMPONENT_INITIAL_PROPERTIES, bundle)
+                }
             }
         }
     }
 
+    private var initialProperties: Bundle? = null
```
ArazAbishov, line 50: As I mentioned above, this prop can be removed too.

PR #91, opened 2020-05-17
refactor(Pods): Allow specifying target platform

### LINKED ISSUE #86: Add support for non-iOS platforms in `test_app.rb`
`use_test_app!` currently assumes target platform is iOS. Allow specifying different targets.

tido64:
Closes #86


macos/test_app.rb
```diff
@@ -0,0 +1,14 @@
...
+# LICENSE file in the root directory of this source tree.
+#
+
+require_relative('../ios/test_app.rb')
+
+def use_test_app!
+  use_test_app_internal!(:macos) do |target|
+    yield target if block_given?
+  end
+end
```
alloy, line 14: FYI You can also capture and pass on a block like so:

```ruby
def use_test_app!(&block)
  use_test_app_internal!(:macOS, &block)
end
```

test/test_test_app.rb
```diff
@@ -19,37 +19,49 @@ def fixture_path(*args)
 
 class TestTestApp < Minitest::Test
   def test_resources_pod_returns_spec_path
-    assert_nil(resources_pod(Pathname.new('/')))
-    assert_nil(resources_pod(Pathname.new('.')))
+    %i[ios macos].each do |target|
```
alloy, line 22: Sometimes it may be preferable to move enumeration of test variants outside of the test and generate multiple tests instead, so you have a better understanding of what failed when an assertion fails. E.g.:

```ruby
class TestTestApp < Minitest::Test
  %i[ios macos].each do |target|
    define_method("test_#{target}_resources_pod_returns_spec_path") do
      assert_nil(resources_pod(Pathname.new('/'), target))
      # ...
    end
  end
end
```

alloy: I guess you could also create `Minitest::Test` subclasses in a loop instead, unsure off-hand what’s more helpful with minitest.

tido64: Thanks, I'll go with the first suggestion for now to keep it simple.

macos/test_app.rb
tido64: Wouldn't this also make `block` a required parameter? I tried this originally but couldn't get it to work properly with `yield` inside of `use_test_app_internal!`.

alloy: It shouldn’t, although my experience might be outdated. Will verify when I get back to my machine, but please feel free to merge in the meantime 👍

alloy: Seems to work, on 2.6 at least:

```
irb(main):001:0> def internal
irb(main):002:1>   yield :ok if block_given?
irb(main):003:1> end
=> :internal
irb(main):004:0> def external(&block)
irb(main):005:1>   internal(&block)
irb(main):006:1> end
=> :external
irb(main):007:0> external
=> nil
irb(main):008:0> external { |x| p x }
:ok
=> :ok
```

tido64: This is what I'm seeing :/

```
% ruby --version
ruby 2.7.1p83 (2020-03-31 revision a0c7c23c9c) [x86_64-darwin19]
% pod install

[!] Invalid `Podfile` file: no block given (yield).

 #  from /~/react-native-test-app/example/ios/Podfile:5
 #  -------------------------------------------
 #
 >  use_test_app!
 #  -------------------------------------------
```

The diff:

```diff
diff --git a/example/ios/Podfile b/example/ios/Podfile
index 59f8153..503f32a 100644
--- a/example/ios/Podfile
+++ b/example/ios/Podfile
@@ -2,8 +2,4 @@ require_relative '../node_modules/react-native-test-app/test_app.rb'

 workspace 'Example.xcworkspace'

-use_test_app! do |target|
-  target.tests do
-    pod 'Example-Tests', :path => '..'
-  end
-end
+use_test_app!
diff --git a/test_app.rb b/test_app.rb
index a1b3ba6..e45d559 100644
--- a/test_app.rb
+++ b/test_app.rb
@@ -7,8 +7,6 @@

 require_relative('ios/test_app.rb')

-def use_test_app!
-  use_test_app_internal!(:ios) do |target|
-    yield target if block_given?
-  end
+def use_test_app!(&block)
+  use_test_app_internal!(:ios, &block)
 end
```

alloy: I think that’s because you also removed this, no?

```
use_test_app! do |target|
  target.tests do
    pod 'Example-Tests', :path => '..'
  end
end
```

tido64: The block passed to `use_test_app!` should be optional, otherwise we'd end up with users having to do something like
```ruby
use_test_app! do |target|
end
```
Hence why I asked whether the block can still be optional. Sorry for not being clearer.

alloy: Ah right, no that’s my bad; you did say so and that’s also why I tested it in my snippet. I think I understand the problem, you should move the `if block_given?` check into `use_test_app_internal` https://github.com/microsoft/react-native-test-app/blob/1a39ae21142d5b05b8f9b51e1fa6abdd9c30cc9b/ios/test_app.rb#L139

tido64: Ahh, right… I forgot that I moved it out. Seems to work now. No idea what issue I hit earlier :/

Anyway: #93

PR #98, opened 2020-05-30
macOS: Add Plop template

### LINKED ISSUE #88: Add macOS target to Plop template
Requires #87

- [x] Add template macOS targets to CI

tido64:
Resolves #88.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -159,6 +159,30 @@ jobs:
           pod install --project-directory=macos
           ../scripts/xcodebuild.sh macos/Example.xcworkspace build
         working-directory: example
+  macos-template:
+    name: "macOS [template]"
+    strategy:
+      matrix:
+        template: [all, macos]
+    runs-on: macos-latest
+    if: "!contains(github.event.head_commit.message, '[skip ci]')"
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install
+        run: scripts/install-test-template.sh ${{ matrix.template }}
+      - name: Build
+        run: |
+          set -eo pipefail
+          yarn build:macos
+          if [[ ${{ matrix.template }} == macos ]]; then
+            pod install
+            ../scripts/xcodebuild.sh TemplateExample.xcworkspace build
+          else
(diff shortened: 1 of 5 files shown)


alloy:
This isn’t part of this PR, but why should the user install Plop separately? https://github.com/microsoft/react-native-test-app/blob/2a1cfd6f268cd1539601c035a3ca45cb89aa8d59/README.md#quick-start

tido64:
> 
> 
> This isn’t part of this PR, but why should the user install Plop separately? https://github.com/microsoft/react-native-test-app/blob/2a1cfd6f268cd1539601c035a3ca45cb89aa8d59/README.md#quick-start

Tbh, I'm just following the documentation: https://plopjs.com/documentation/#installation. But on closer inspection, that may be for template devs and not for consumers. I'll check if using `npx` is enough, and update the instructions.

alloy:
Noticing this deprecation warning, is that on your radar?

```
[!] Passing custom root to use_native_modules! is deprecated.
    - CLI detects root of the project automatically. The "." argument was ignored.
```

tido64:
> Noticing this deprecation warning, is that on your radar?
> 
> ```
> [!] Passing custom root to use_native_modules! is deprecated.
>     - CLI detects root of the project automatically. The "." argument was ignored.
> ```

Yup, addressed here: https://github.com/microsoft/react-native-test-app/pull/17/files#diff-ccb88c6fbd1a6787a22c613fbff1b980

alloy:
> Tbh, I'm just following the documentation: https://plopjs.com/documentation/#installation. But on closer inspection, that may be for template devs and not for consumers. I'll check if using `npx` is enough, and update the instructions.

It’s not immediately clear to me what scenario they describe there, but for our case making Plop a runtime dependency and instructing users to install `react-native-test-app` as a dev dependency should be enough imo.

alloy:
Was able to build the macOS app, given the included instructions :+1:

tido64:
> Was able to build the macOS app, given the included instructions +1

Thanks for the thorough review :heart:

alloy:
You got it, buddy!

PR #99, opened 2020-06-04
fix: Don't require Plop to be installed separately

tido64:


alloy:
Hmm, why use `npx`? I think that if you move `plop` in `package.json` to `dependencies` would be a better experience. That way the instructions would be:

```
yarn add react-native-test-app --dev
yarn plop --plopfile node_modules/react-native-test-app/plopfile.js --dest sample
```

tido64:
Aha, that's a good point. Didn't think of that.

alloy:
Great! 👍

PR #102, opened 2020-06-06
fix: `pod install` fails in monorepos using React Native 0.60

tido64:


alloy:
Nothing really stands out to me, but it’s a little hard to judge the solution when the problem is only described in the abstract. If it works for you then I would say merge it 👍

One thing to keep in mind is wether or not some of these issues are specific to this project or should perhaps be fixed upstream so it works for other monorepos too.

tido64:
> One thing to keep in mind is wether or not some of these issues are specific to this project or should perhaps be fixed upstream so it works for other monorepos too.

The issue is that React Native 0.60 depends on react-native-cli 2.x, which doesn't fully support monorepos. We need to specify the project path, `'.'`, which was deprecated in 3.x. So instead of adding it unconditionally like we did prior to the RN bump to 0.61, we check the version so we don't get the deprecation warning.

alloy:
Aha, got it; makes sense! 👌

PR #104, opened 2020-06-09
fix(macOS): Present the component immediately if there's only one

tido64:
rnta-macos-start

### DIFF
--- a/macos/ReactTestApp/AppDelegate.swift
+++ b/macos/ReactTestApp/AppDelegate.swift
@@ -9,78 +9,85 @@ import Cocoa
 
 @NSApplicationMain
 final class AppDelegate: NSObject, NSApplicationDelegate {
+    @IBOutlet weak var reactMenu: NSMenu!
+
     private(set) lazy var reactInstance = ReactInstance()
+
+    private weak var mainWindow: NSWindow?
 
     func applicationShouldTerminateAfterLastWindowClosed(_ sender: NSApplication) -> Bool {
         true
     }
 
     func applicationDidFinishLaunching(_ aNotification: Notification) {
-        guard let mainMenu = NSApplication.shared.mainMenu else {
-            return
-        }
+        // `keyWindow` might be `nil` while loading or when the window is not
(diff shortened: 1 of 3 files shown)


macos/ReactTestApp/AppDelegate.swift
```diff
@@ -11,6 +11,13 @@ import Cocoa
 final class AppDelegate: NSObject, NSApplicationDelegate {
     private(set) lazy var reactInstance = ReactInstance()
 
+    private var mainWindow: NSWindow? {
+        // `keyWindow` might be `nil` while loading or when the window is not
+        // active. Use `frameAutosaveName` to find our main window.
+        let windows = NSApplication.shared.windows
+        return windows.first { $0.frameAutosaveName == "MainWindow" }
+    }
```
alloy, line 19: It seems like a better idea to either hookup the window from the nib/storyboard to the `AppDelegate`; _or_, if the nib/storyboard doesn’t do anything of interest, just remove the window from there and create it programmatically instead.

macos/ReactTestApp/AppDelegate.swift
```diff
@@ -63,7 +73,9 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
             item.isEnabled = false
         }
 
+        // 0             1     2     3       4     5
         // ReactTestApp  File  Edit  Format  View  React  Window  Help
+        //                                         ^~~~~ Insert `reactMenu` here
```
alloy, line 78: Similarly to my above comment, it might be more consistent to instantiate the menu in your nib/storyboard, like all the other existing menu items; _or_ create all of them programmatically, which seems like overkill–there may also be a way to insert an item by means of ‘menu validation’, but it’s been too long for me to remember off hand.

macos/ReactTestApp/AppDelegate.swift
alloy: Sounds good, be sure to file an issue and link it here please 🙏

tido64: Sure thing: #111

PR #105, opened 2020-06-09
fix(iOS): Present the component immediately if there's only one

tido64:
ios-launch


sweggersen:
Do we have a ticket for other platforms?

tido64:
> 
> 
> Do we have a ticket for other platforms?

Not really. I was just going to do Android as well (unless you want to :smile:)

PR #106, opened 2020-06-09
fix(Android): Present the component immediately if there's only one

tido64:
android-startup

### DIFF
--- a/android/app/src/main/java/com/sample/MainActivity.kt
+++ b/android/app/src/main/java/com/sample/MainActivity.kt
@@ -31,12 +31,22 @@ class MainActivity : ReactActivity() {
     private val testAppReactNativeHost: TestAppReactNativeHost
         get() = reactNativeHost as TestAppReactNativeHost
 
-    private val listener = { component: ComponentViewModel ->
-        startActivity(
-            ComponentActivity.newIntent(
-                this, component.name, component.displayName, component.initialProperties
-            )
+    private val newComponentActivityIntent = { component: ComponentViewModel ->
+        ComponentActivity.newIntent(
+            this, component.name, component.displayName, component.initialProperties
         )
+    }
+
+    private val newComponentViewModel = { component: Component ->
+        ComponentViewModel(
+            component.appKey,
+            component.displayName ?: component.appKey,
+            component.initialProperties
+        )
+    }
+
+    private val startComponentActivity = { component: ComponentViewModel ->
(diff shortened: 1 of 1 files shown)


ArazAbishov:
Hmm, you will lose access to the dev menu options on start. Also, backstack behaviour will be somewhat unintuitive where it gets you back to a screen that you haven't seen when launching the app. 

It makes a lot of sense on desktop where you land to a component on start and always have access to contextual menus. If we want to make it "cleaner" on mobile I would suggest to make dev menu somehow accessible everywhere and do not allow to go back to the list of components.

ArazAbishov:
> The issue I hit next was that we'd always load the bundle from disk, so I turned it on as long as it is a debug build.
Ah, I see. I would rather keep the initial implementation then. 

Do you think it will make sense to implement a different navigation pattern where having a default component feels natural? Something like a right hand side sliding drawer? For example:

 

Obviously, this will require more work and thought. We can get the PR in as the first implementation, but I would like us to think of other solutions too. @sweggersen any input on this?

sweggersen:
I personally think that `BuildConfig.DEBUG` is not the correct flag to spot intent, since basically all builds yo try in AS will be DEBUG.

I've had bad experiences when the app tries to be smart and loads the bundle from disk if available. I have to click load from dev server every time.
Can we do something like this:

image

ArazAbishov:
@sweggersen How would you access the settings screen when working with a single experience only?

sweggersen:
Right. My diagram is a bit of on that part. I thought we would launch the single experience Activity automatically, but still launch the experience list first, so that when you exit the experience, you would go back to the experience list where we have control of the toolbar menu, and have it there.

tido64:
On iOS, RN checks if the dev server is running first, then load the embedded bundle as fallback. I think we should implement (and upstream?) that feature. As @sweggersen experienced, I think it's wrong that it tries the embedded bundle first. As our users are developers, it's more likely that they want the dev server most of the time. I've created the issue here: #115. Let's revisit this after it is resolved.

tido64:
> Do you think it will make sense to implement a different navigation pattern where having a default component feels natural?

@ArazAbishov: Can we simply keep the current toolbar? We only have a few menu items, I'm not sure we need a whole right panel. In any case, it would be nice to get this in first and create a separate issue for it.

tido64:
Opened #117

PR #120, opened 2020-06-16
fix(macOS): macOS target should be optional

tido64:
Projects that don't care for `react-native-macos` shouldn't have to install it just to run `plop`. Now we should get a warning instead, and none of the macOS specific commands/files:

```
% yarn plop --dest template-example TemplateExample all
yarn run v1.22.4
⠋ [WARN] Cannot find module 'react-native-macos'; skipping macOS target
✔  ++ /.watchmanconfig
✔  ++ /App.js
✔  ++ /app.json
✔  ++ /babel.config.js
✔  ++ /index.js
✔  ++ /metro.config.js
✔  ++ /package.json
✔  ++ /react-native.config.js
✔  ++ /android/gradle/wrapper/gradle-wrapper.jar
✔  ++ /android/gradle/wrapper/gradle-wrapper.properties
✔  ++ /android/gradle.properties
✔  ++ /android/gradlew
✔  ++ /android/gradlew.bat
✔  ++ /android/settings.gradle
✔  ++ /ios/Podfile
```

### DIFF
--- a/plopfile.js
+++ b/plopfile.js
@@ -10,6 +10,25 @@
  * }} InputData
  */
 
+const chalk = require("chalk");
+
+/**
+ * Returns whether the specified package is installed.
+ * @param {string} pkg The target package, e.g. "react-native-macos"
+ * @param {boolean} isRequired Whether the package is required
+ * @return {boolean}
+ */
+function isInstalled(pkg, isRequired) {
+  try {
+    return Boolean(require.resolve(pkg));
+  } catch (error) {
+    if (isRequired) {
+      throw error;
+    }
+    return false;
+  }
+}
+
 /**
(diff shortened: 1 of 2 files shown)


plopfile.js
```diff
@@ -50,7 +69,12 @@ module.exports = (plop) => {
     ],
     actions: (/** @type {InputData} */ { name, platforms }) => {
       const path = require("path");
+
       const exclusive = platforms !== "all";
+      const includeMacOS =
```
tom-un, line 74: Shouldn't there be a similar check to see if react-native-windows is installed or not?

tido64: When we get Windows support, we will have to add it. For now, I think this is sufficient.

PR #121, opened 2020-06-16
fix: Allow users to enable Flipper

### LINKED ISSUE #112: Add support for enabling Flipper
Give users a way to enable Flipper when on React Native 0.62

tido64:
Resolves #112

To test these changes, bump to 0.62 and enable Flipper:
```diff
diff --git a/example/android/gradle.properties b/example/android/gradle.properties
index 08bd26e..699258c 100644
--- a/example/android/gradle.properties
+++ b/example/android/gradle.properties
@@ -1,3 +1,5 @@
 # These properties are required to enable AndroidX for the test app.
 android.useAndroidX=true
 android.enableJetifier=true
+
+FLIPPER_VERSION=0.33.1
diff --git a/example/ios/Podfile b/example/ios/Podfile
index 59f8153..512a835 100644
--- a/example/ios/Podfile
+++ b/example/ios/Podfile
@@ -2,6 +2,7 @@ require_relative '../node_modules/react-native-test-app/test_app.rb'

 workspace 'Example.xcworkspace'

+use_flipper!
 use_test_app! do |target|
   target.tests do
     pod 'Example-Tests', :path => '..'
diff --git a/example/package.json b/example/package.json
index f245bf9..80824e0 100644
--- a/example/package.json
+++ b/example/package.json
@@ -17,9 +17,8 @@
   "devDependencies": {
     "@babel/core": "^7.0.0",
     "mkdirp": "^0.5.1",
-    "react": "16.9.0",
-    "react-native": "0.61.5",
-    "react-native-macos": "0.61.39",
+    "react": "16.11.0",
+    "react-native": "0.62.2",
     "react-native-test-app": "../"
   }
 }
```

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-    ext.kotlinVersion = "1.3.70"
+    ext.kotlinVersion = "1.3.72"
 
     def buildscriptDir = buildscript.sourceFile.getParent()
     apply from: "$buildscriptDir/../test-app-util.gradle"
@@ -34,11 +34,21 @@ def testAppDir = file("$projectDir/../../")
 apply from: file("${testAppDir}/test-app.gradle")
 applyTestAppModule(project, "com.sample")
 
-project.ext.react = [enableHermes: true]
+project.ext.react = [
+  enableFlipper: hasProperty('FLIPPER_VERSION'),
+  enableHermes: true,
+]
 
 android {
     compileSdkVersion 29
     buildToolsVersion "29.0.3"
+
+    // TODO: Remove this block when minSdkVersion >= 24. See
+    // https://stackoverflow.com/q/53402639 for details.
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
 
     defaultConfig {
         applicationId "com.sample"
@@ -54,6 +64,12 @@ android {
         pickFirst "lib/arm64-v8a/libc++_shared.so"
         pickFirst "lib/x86_64/libc++_shared.so"
         pickFirst "lib/x86/libc++_shared.so"
+    }
+
+    sourceSets {
+        if (project.ext.react.enableFlipper) {
+            debug.java.srcDirs += 'src/flipper/java'
+        }
     }
 }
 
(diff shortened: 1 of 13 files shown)


android/app/build.gradle
```diff
@@ -55,6 +65,12 @@ android {
         pickFirst "lib/x86_64/libc++_shared.so"
         pickFirst "lib/x86/libc++_shared.so"
```
sweggersen, line 66: You can add 

     buildConfigField 'boolean', 'USE_FLIPPER', project.ext.react.enableFlipper

inside `defaultConfig`. Then you'll be able to use `BuildConfig.USE_FLIPPER` in `TestAppReactNativeHost.kt` instead of `BuildConfig.DEBUG`, and then you can remove the try/catch block :)

sweggersen: Then we also match the logic on iOS

tido64: `enableFlipper` could be turned on but it won't be included in release builds. That's why we use reflection here. Is there a way to have `BuildConfig.USE_FLIPPER` only be true in debug builds?

sweggersen: Yes, okay, so instead of using `defaultConfig` you can add this below:

    android {
       ...
      buildTypes {
        release {
            buildConfigField 'boolean', 'USE_FLIPPER', false
        }

        debug {
            buildConfigField 'boolean', 'USE_FLIPPER', project.ext.react.enableFlipper
        }
      }
      ..
    }

tido64: Hmm... But we still need to `import` on top, right?

sweggersen: Talked offline. Think what you have is fine

sweggersen:
> Is there a reason to not default to setting `use_flipper` to `true`? In the context of a test application, I would assume defaulting to making it as debuggable as possible would be a goal?

I think it's because we need RN 0.62 for it to work.

tido64:
> > Is there a reason to not default to setting `use_flipper` to `true`? In the context of a test application, I would assume defaulting to making it as debuggable as possible would be a goal?
> 
> I think it's because we need RN 0.62 for it to work.

That said, I think it still makes sense to enable it by default if you're on >= 0.62. I've opened an issue here: #124.

Thanks for reviewing, everyone. :heart:

alloy:
Got it, makes sense. Thanks for the follow-up issue! 🙏

PR #126, opened 2020-06-21
fix: Enable Flipper by default when on 0.62+

### LINKED ISSUE #124: Enable Flipper by default when on RN 0.62+
Flipper is disabled by default today to keep backwards compatibility with 0.60. We should try to detect whether the user is on 0.62 and enable it by default.

- [ ] Detect whether we're on RN 0.62 or greater and enable Flipper by default
- [ ] Update [documentation](https://github.com/microsoft/react-native-test-app/wiki#enable-flipper) on how to **opt out** of Flipper

tido64:
Resolves #124

### Test Plan

1. On React Native 0.60 and 0.61, Flipper should remain off even if it's been explicitly turned on
2. On React Native 0.62, Flipper should be enabled with the recommended versions if unspecified
3. On React Native 0.62, enabling Flipper with specific versions should override the recommended versions
4. On React Native 0.62, it is possible to disable Flipper by specifying `false`, i.e. `FLIPPER_VERSION=false` or `use_flipper!(false)`

### DIFF
--- a/android/test-app-util.gradle
+++ b/android/test-app-util.gradle
@@ -51,3 +51,25 @@ ext.findNodeModulesPath = { baseDir, packageName ->
 
     return null
 }
+
+ext.getFlipperVersion = { baseDir ->
+    def reactNativePath = findNodeModulesPath(baseDir, 'react-native')
+    def props = new Properties()
+    file("${reactNativePath}/template/android/gradle.properties").withInputStream {
+        props.load(it)
+    }
+
+    def recommendedFlipperVersion = props.getProperty('FLIPPER_VERSION')
+    if (recommendedFlipperVersion == null) {
+        // Current React Native version doesn't support Flipper
+        return null
+    }
+
+    // Prefer user specified Flipper version
+    if (project.hasProperty('FLIPPER_VERSION')) {
+        return project.getProperty('FLIPPER_VERSION')
+    }
+
+    // Use the recommended Flipper version
+    return recommendedFlipperVersion
+}
--- a/ios/test_app.rb
+++ b/ios/test_app.rb
@@ -8,8 +8,6 @@
 require('json')
 require('pathname')
 
-@use_flipper = false
-
 def assert(condition, message)
   raise message unless condition
 end
@@ -40,6 +38,10 @@ def find_project_root
   assert(!podfile_path.nil?, "Could not find 'Podfile'")
 
   Pathname.new(File.dirname(podfile_path.absolute_path))
+end
+
+def flipper_enabled?(react_native_version)
+  react_native_version >= 6200 && @flipper_versions != false
 end
 
 def nearest_node_modules(project_root)
@@ -96,7 +98,7 @@ def resources_pod(project_root, target_platform)
 end
 
 def use_flipper!(versions = {})
-  @use_flipper = versions
+  @flipper_versions = versions
 end
 
 def use_react_native!(project_root, target_platform)
@@ -116,7 +118,7 @@ def use_react_native!(project_root, target_platform)
   include_react_native!(react_native: react_native.relative_path_from(project_root).to_s,
                         target_platform: target_platform,
                         project_root: project_root,
-                        use_flipper: @use_flipper)
+                        flipper_versions: @flipper_versions || {})
 end
 
 def make_project!(xcodeproj, project_root, target_platform)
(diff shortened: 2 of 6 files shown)


android/test-app-util.gradle
```diff
@@ -51,3 +51,25 @@ ext.findNodeModulesPath = { baseDir, packageName ->
...
+        return null
+    }
+
+    // Prefer user specified Flipper version
+    if (project.hasProperty('FLIPPER_VERSION')) {
+        return project.getProperty('FLIPPER_VERSION')
+    }
+
+    // Use the recommended Flipper version
+    return flipperVersion
```
sweggersen, line 74: We account for both 'FLIPPER_VERSION' being null or having a value before, so would we ever get here?

tido64: There's a `FLIPPER_VERSION` which the user sets in their `gradle.properties`, and there's another one in `react-native` template that we use to detect whether the current version supports Flipper at all, as well as getting the recommended version. It's a bit confusing since we have to check that the value exists, and if so, check if the user has defined a version before returning the value we just got. Maybe it helps if I rename the variable to `recommendedFlipperVersion`?

sweggersen: Aha, I see. I got confused there. Yeah maybe that would help the next person reading this.

ios/test_app.rb
```diff
@@ -116,7 +114,7 @@ def use_react_native!(project_root, target_platform)
   include_react_native!(react_native: react_native.relative_path_from(project_root).to_s,
                         target_platform: target_platform,
                         project_root: project_root,
-                        use_flipper: @use_flipper)
+                        use_flipper: @use_flipper.nil? ? {} : @use_flipper)
```
alloy, line 117: I don’t quite understand the intent here. Passing `{}` means the `if target_platform == :ios && use_flipper` check in `include_react_native!` will pass if `@use_flipper` is not explicitly set. If that’s the case, why not use `true` instead of `{}`? And additionally, how does this handle the conditional default enabling outlined in your PR description?

tido64: I think this will make more sense if we rename `@use_flipper` -> `@flipper_versions`. The parameter is forwarded to `add_flipper_pods!` in `ios/use_react_native-0.62.rb`. This parameter is not used in 0.60 and 0.61, so Flipper wouldn't be enabled. In 0.63, it will be forwarded to `use_flipper!` from upstream.

```ruby
def add_flipper_pods!(versions = {})
  versions['Flipper'] ||= '~> 0.33.1'
  versions['DoubleConversion'] ||= '1.1.7'
  versions['Flipper-Folly'] ||= '~> 2.1'
  versions['Flipper-Glog'] ||= '0.3.6'
  versions['Flipper-PeerTalk'] ||= '~> 0.0.4'
  versions['Flipper-RSocket'] ||= '~> 1.0'
  ...
```

ios/test_app.rb
```diff
@@ -116,7 +114,7 @@ def use_react_native!(project_root, target_platform)
   include_react_native!(react_native: react_native.relative_path_from(project_root).to_s,
                         target_platform: target_platform,
                         project_root: project_root,
-                        use_flipper: @use_flipper)
+                        flipper_versions: @flipper_versions.nil? ? {} : @flipper_versions)
```
alloy, line 117: Mostly a idiomatic Ruby FYI:

```suggestion
                        flipper_versions: @flipper_versions || {})
```

ios/test_app.rb
```diff
@@ -151,7 +149,7 @@ def make_project!(xcodeproj, project_root, target_platform)
     next if target.name != 'ReactTestApp'
 
     target.build_configurations.each do |config|
-      use_flipper = config.name == 'Debug' && @use_flipper
+      use_flipper = config.name == 'Debug' && @flipper_versions != false
```
alloy, line 152: Seems like testing against a boolean here is not correct any longer? Perhaps just check for any falsy value?

```suggestion
      use_flipper = config.name == 'Debug' && @flipper_versions
```

ios/test_app.rb
alloy: Ahh, gotcha 👍

ios/test_app.rb
tido64: I think we still need to compare against `false` since `nil` should be to enable by default:

- `@flipper_versions == nil`: Enable Flipper if on RN 0.62+
- `@flipper_versions == { ... }`: Enable Flipper with specified versions  if on RN 0.62+
- `@flipper_versions == false`: Disable Flipper

I've pulled this out into a function. What are your thoughts?

tido64:
> Looks good, thanks for bearing with me! 👌

@alloy: Thanks for being thorough :heart:

PR #127, opened 2020-06-21
Add support for 0.63

tido64:
### Test Plan

1. Apply the patch below
2. Run the usual install commands (i.e. `yarn`, `pod install`, etc.)
3. Build the JS bundle, i.e. `yarn build:android` and `yarn build:ios`
4. Build the app in Android Studio or Xcode

```diff
diff --git a/example/package.json b/example/package.json
index e16f729..645099c 100644
--- a/example/package.json
+++ b/example/package.json
@@ -17,9 +17,8 @@
   "devDependencies": {
     "@babel/core": "^7.0.0",
     "mkdirp": "^0.5.1",
-    "react": "16.9.0",
-    "react-native": "0.61.5",
-    "react-native-macos": "0.61.39",
+    "react": "16.13.1",
+    "react-native": "0.63.0-rc.1",
     "react-native-test-app": "../"
   }
 }
```


ios/test_app.rb
```diff
@@ -53,7 +53,12 @@ def nearest_node_modules(project_root)
 
 def package_version(package_path)
   package_json = JSON.parse(File.read(File.join(package_path, 'package.json')))
-  package_json['version'].match(/(?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)/)
+  match = package_json['version'].match(/(?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)/)
+
+  # Convert #<MatchData "0.62.2" major:"0" minor:"62" patch:"2">
+  #      => {:major=>0, :minor=>62, :patch=>2}
+  # Source: https://gist.github.com/bchase/d9ecfc1d58e38f158793525864583b17
+  [match.names.map(&:to_sym), match.captures.map(&:to_i)].transpose.to_h
```
alloy, line 61: The Ruby stdlib should include RubyGems iirc, which means you may just as well require its `Version` class: https://github.com/rubygems/rubygems/blob/master/lib/rubygems/version.rb

ios/use_react_native-0.63.rb
```diff
@@ -0,0 +1,12 @@
+#
+# Copyright (c) Microsoft Corporation
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+#
+
+def include_react_native!(react_native:, target_platform:, project_root:, flipper_versions:)
+  require_relative "#{File.join(project_root, react_native)}/scripts/react_native_pods"
```
alloy, line 9:
```suggestion
  require_relative File.join(project_root, react_native, "scripts", "react_native_pods")
```

ios/test_app.rb
tido64: Is that something that needs to be installed separately? I don't see it in `gem list`.

alloy: No, because stdlib libraries won’t show up in `gem list`:

```
irb(main):001:0> require "rubygems/version"
=> false
irb(main):002:0> Gem::Version.new("0.63.0-rc.1") < Gem::Version.new("0.63.0-rc.2")
=> true
irb(main):003:0> Gem::Version.new("0.63.0-rc.1") < Gem::Version.new("0.63.0-rc.0")
=> false
```

tido64: Thanks. This looks nicer.

ios/test_app.rb
```diff
@@ -53,7 +53,8 @@ def nearest_node_modules(project_root)
 
 def package_version(package_path)
   package_json = JSON.parse(File.read(File.join(package_path, 'package.json')))
-  package_json['version'].match(/(?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)/)
+  match = package_json['version'].match(/(?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)/)
+  Gem::Version.new(match[0])
```
alloy, line 57: Why is there still a need to match at all, instead of `Gem::Version.new(package_json['version'])`?

tido64: We don't. I just missed it :smile:

alloy: Hah, perfect 👍

PR #133, opened 2020-06-25
fix: Wrong pods get installed with release candidates

tido64:


tido64:
@alloy: I'm having issues adding tests for `resolve_module` (and all the functions calling it) because it makes a call to `Pod::Executable.execute_command`. I'm reluctant to replace it just so we can add tests. Any ideas?

alloy:
@tido64 minitest comes with [mocking](https://github.com/seattlerb/minitest#mocks-) and [stubbing](https://github.com/seattlerb/minitest#stubs-) facilities.

PR #134, opened 2020-06-25
test: Add tests for autolink_script_path, autolink_script_version

tido64:


test/test_test_app.rb
```diff
@@ -18,6 +18,20 @@ def fixture_path(*args)
 end
 
 class TestTestApp < Minitest::Test
+  def test_autolink_script_path
+    cli = fixture_path('test_app', 'node_modules', '@react-native-community', 'cli-platform-ios')
+    stub :resolve_module, cli do
```
alloy, line 23: Just an FYI, but top-level methods exist on the `Kernel` module, so it _might_ read a bit more clear if it was explicitly written as:

```suggestion
    Kernel.stub :resolve_module, cli do
```

Not in any way a blocker, though.

alloy: In fact, perhaps it might be cleaner to move all those top-level methods into a RNTA namespace module? 🤔 

```ruby
module RNTA
  def self.resolve_module(…)
  end
end
```

Anyways, for another PR.

tido64: Are you sure about the `Kernel` module? Doesn't look like it stubbed the correct function:
```
% bundle exec ruby -Ilib:test test/test_test_app.rb
Run options: --seed 56425

# Running:

...E..E....

Finished in 0.016080s, 684.0796 runs/s, 3544.7761 assertions/s.

  1) Error:
TestTestApp#test_autolink_script_path:
NameError: uninitialized constant Pod
    /~/react-native-test-app/ios/test_app.rb:80:in `resolve_module'
    /~/react-native-test-app/ios/test_app.rb:17:in `autolink_script_path'
    test/test_test_app.rb:24:in `block in test_autolink_script_path'

  2) Error:
TestTestApp#test_autolink_script_version:
NameError: uninitialized constant Pod
    /~/react-native-test-app/ios/test_app.rb:80:in `resolve_module'
    /~/react-native-test-app/ios/test_app.rb:22:in `autolink_script_version'
    test/test_test_app.rb:31:in `block in test_autolink_script_version'

11 runs, 57 assertions, 0 failures, 2 errors, 0 skips
```

tido64: I'll merge this for now and look into moving stuff into a module.

alloy: Interesting. It looks like it does stub the right method, but then lookup of the `Pod` constant fails 🤔

Issue #34, opened 2020-03-25
Give Android files a proper namespace


tido64:
The Android files currently live under `com.sample`. Let's move it into a proper space, e.g. `com.react.testapp`.

PR #137, opened 2020-06-30
fix(android): rename the root package from com.sample to com.react.testapp

ArazAbishov:
Renames the root package and its references from com.sample to com.react.testapp #34


android/app/src/main/java/com/react/testapp/MainActivity.kt
```diff
@@ -117,7 +118,7 @@ class MainActivity : ReactActivity() {
 
         findViewById<TextView>(R.id.runtime_info).apply {
             text = resources.getString(
-                R.string.runtime_info,
+                    R.string.runtime_info,
```
tido64, line 121: Intentional indent?
```suggestion
                R.string.runtime_info,
```

PR #138, opened 2020-07-01
fix: Always copy 'app.json'

tido64:


ios/test_app.rb
```diff
@@ -91,8 +91,8 @@ def resources_pod(project_root, target_platform)
   app_manifest = find_file('app.json', project_root)
   return if app_manifest.nil?
 
-  resources = resolve_resources(JSON.parse(File.read(app_manifest)), target_platform)
-  return if !resources.instance_of?(Array) || resources.empty?
+  manifest_resources = resolve_resources(JSON.parse(File.read(app_manifest)), target_platform)
+  resources = ['app.json'] + (manifest_resources.instance_of?(Array) ? manifest_resources : [])
```
alloy, line 95: What else could `manifest_resources` be and should it warrant throwing an error?

tido64: It could be `nil`, e.g. because it is missing in the manifest, and that is fine.

alloy: Gotcha. Just FYI, a more idiomatic way could be to use the ‘splat’ operator:

```
irb(main):001:0> ["app.json", *nil]
=> ["app.json"]
irb(main):002:0> ["app.json", *["foo"]]
=> ["app.json", "foo"]
```

tido64: Ooh, nice. I'll try it out next time I submit a PR 😄

PR #139, opened 2020-07-01
fix(iOS): Add URI schemes to support MSAL

tido64:
As documented here: https://github.com/AzureAD/microsoft-authentication-library-for-objc#ios-only-steps

### DIFF
--- a/ios/ReactTestApp/Info.plist
+++ b/ios/ReactTestApp/Info.plist
@@ -16,6 +16,15 @@
 	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
 	<key>CFBundleShortVersionString</key>
 	<string>1.0</string>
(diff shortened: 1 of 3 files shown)


ios/ReactTestApp/Info.plist
```diff
@@ -16,6 +16,15 @@
 	<string>$(PRODUCT_BUNDLE_PACKAGE_TYPE)</string>
 	<key>CFBundleShortVersionString</key>
 	<string>1.0</string>
+	<key>CFBundleURLTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>msauth.$(PRODUCT_BUNDLE_IDENTIFIER)</string>
```
tido64, line 24: We need a way for folks to change `PRODUCT_BUNDLE_IDENTIFIER`.

PR #141, opened 2020-07-02
fix(iOS): Split out Xcode build configuration

### LINKED ISSUE #130: Split out Xcode build configuration to separate files
[Xcode build configuration files](https://nshipster.com/xcconfig/) allow users to manage build configurations outside of Xcode. There is currently no obvious way to include them in `react-native-test-app`.

Example project with `.xcconfig` files: https://github.com/microsoft/fluentui-react-native/tree/master/apps/ios/src

cc @alloy

tido64:
Resolves #130


ios/test_app.rb
```diff
@@ -142,15 +144,37 @@ def make_project!(xcodeproj, project_root, target_platform)
...
+
+      dst_xcconfig = File.expand_path("ReactTestApp.#{config}.xcconfig", destination)
+      File.open(dst_xcconfig, 'w') do |fo|
+        pods_xcconfig_path = File.join(
+          project_root.relative_path_from(destination).to_s,
+          'Pods',
+          'Target Support Files',
+          'Pods-ReactTestApp',
+          "Pods-ReactTestApp.#{config}.xcconfig"
+        )
```
tido64, line 163: @alloy: Is this path stable? Is there a better way to get to the `.xcconfig` files generated by CocoaPods?

alloy: Yeah that should be stable enough 👍

ios/test_app.rb
```diff
@@ -142,15 +144,24 @@ def make_project!(xcodeproj, project_root, target_platform)
   FileUtils.cp(File.join(src_xcodeproj, 'project.pbxproj'), dst_xcodeproj)
   FileUtils.ln_sf(File.join(src_xcodeproj, 'xcshareddata'), dst_xcodeproj)
 
+  # Link Xcode build configuration files
```
tido64, line 147: If we no longer need to modify them, we can just move them into `ReactTestApp` folder instead. `ReactTestApp` gets linked below and we'll no longer need this block.

alloy: Ah, even better 👍

ios/test_app.rb
```diff
@@ -106,7 +108,7 @@ def resources_pod(project_root, target_platform)
       'ios' => '12.0',
       'osx' => '10.14',
     },
-    'resources' => resources,
+    'resources' => ['app.json', *resources],
```
alloy, line 111: 👌 ✨

PR #142, opened 2020-07-02
Load components from manifest

AnastasiaOrishchenko:


windows/ReactTestApp/Manifest.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include <list>
+#include <map>
+#include <optional>
+#include <string>
+
+#include <nlohmann/json.hpp>
+
+using json = nlohmann::json;
+using namespace std;
```
tido64, line 11: Same here. No `use namespace` or `using` in header files.

windows/ReactTestApp/Manifest.h
```diff
@@ -0,0 +1,52 @@
...
+        c.displayName = get_optional<string>(j, "displayName");
+        c.initialProperties = get_optional<map<string, string>>(j, "initialProperties");
+    }
+
+    inline void from_json(const json &j, Manifest &m)
+    {
+        m.name = j.at("name");
+        m.displayName = j.at("displayName");
+        m.components = j.at("components").get<list<Component>>();
+    }
```
tido64, line 47: Can we move these to `.cpp`?

windows/ReactTestApp/Manifest.h
```diff
@@ -0,0 +1,52 @@
...
+    struct Component {
+        string appKey;
+        std::optional<string> displayName;
+        std::optional<map<string, string>> initialProperties;
+    };
+
+    struct Manifest {
+        string name;
+        string displayName;
+        list<Component> components;
```
tido64, line 23: Can we use `std::vector` instead?

windows/ReactTestApp/Manifest.h
```diff
@@ -0,0 +1,52 @@
...
+#include <map>
+#include <optional>
+#include <string>
+
+#include <nlohmann/json.hpp>
+
+using json = nlohmann::json;
+using namespace std;
+
+namespace winrt::ReactTestApp::implementation {
```
tido64, line 13: This should just be in its own namespace.
```suggestion
namespace ReactTestApp
 {
```

windows/ReactTestApp/Manifest.h
```diff
@@ -0,0 +1,52 @@
...
+    inline void from_json(const json &j, Manifest &m)
+    {
+        m.name = j.at("name");
+        m.displayName = j.at("displayName");
+        m.components = j.at("components").get<list<Component>>();
+    }
+
+    struct ManifestProvider {
+        static Manifest getManifest();
+    };
```
tido64, line 51: Prefer free functions over static class functions.

windows/ReactTestApp/Manifest.h
```diff
@@ -0,0 +1,52 @@
...
+
+    template <typename T>
+    std::optional<T> get_optional(const json &j, const string &key)
+    try {
+        return j.at(key).get<T>();
+    }
+    catch (const json::exception &) {
+        return std::nullopt;
+    }
```
tido64, line 33: This block looks like it's missing `{ }`

windows/ReactTestApp/Manifest.h
```diff
@@ -0,0 +1,52 @@
...
+    };
+
+    struct Manifest {
+        string name;
+        string displayName;
+        list<Component> components;
+    };
+
+    template <typename T>
+    std::optional<T> get_optional(const json &j, const string &key)
```
tido64, line 27:
```suggestion
    std::optional<T> try_get(const nlohmann::json &json, const std::string &key)
```

windows/ReactTestApp/Manifest.h
```diff
@@ -0,0 +1,52 @@
...
+    try {
+        return j.at(key).get<T>();
+    }
+    catch (const json::exception &) {
```
tido64, line 31: Does nlohmann/json have APIs that are exception free? We should use those if available.

windows/ReactTestApp/Manifest.h
```diff
@@ -0,0 +1,21 @@
...
+    };
+
+    struct Manifest {
+        std::string name;
+        std::string displayName;
+        std::vector<Component> components;
+    };
+
+    Manifest getManifest();
+
```
tido64, line 20: Nit: Trim newline.
```suggestion
```

windows/ReactTestApp/Manifest.h
```diff
@@ -0,0 +1,21 @@
...
+        std::optional<std::map<std::string, std::string>> initialProperties;
+    };
+
+    struct Manifest {
+        std::string name;
+        std::string displayName;
+        std::vector<Component> components;
+    };
+
+    Manifest getManifest();
```
tido64, line 19: Seems the naming convention in a UWP app is CamelCase, so we should use `GetManifest()` here.
```suggestion
    Manifest GetManifest();
```

windows/ReactTestApp/Manifest.cpp
```diff
@@ -0,0 +1,42 @@
...
+{
+    template <typename T>
+    std::optional<T> try_get(const nlohmann::json &j, const std::string &key)
+    {
+        try {
+            return j.at(key).get<T>();
+        } catch (const nlohmann::json::exception &) {
+            return std::nullopt;
+        }
+    }  // namespace ReactTestApp
```
tido64, line 18: This comment no longer applies
```suggestion
    }
```

windows/ReactTestApp/Manifest.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include <nlohmann/json.hpp>
```
tido64, line 3: We can move `#include <nlohmann/json.hpp>` to the `.cpp` file since it is unused here. Then add the headers we do need:
```suggestion
#include <map>
#include <optional>
#include <string>
#include <vector>
```

windows/ReactTestApp/Manifest.h
AnastasiaOrishchenko: Changed it now. Not sure if it's the best possible solution. but probably better than catching exception

windows/ReactTestApp/Manifest.h
```diff
@@ -0,0 +1,24 @@
...
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace winrt::ReactTestApp::implementation
+{
+    struct Component {
+        std::string appKey;
+        std::optional<std::string> displayName;
+        std::optional<std::map<std::string, std::string>> initialProperties;
```
tido64, line 13: It looks like this only supports string values. Initial properties supports other types as well though, as reported in #144. I don't think we should do anything now in this PR but we will have to revisit this when integrating React Native. It looks like we'll need to convert this to [`folly::dynamic`](https://github.com/microsoft/react-native-windows/blob/532e299688d745c0d7c5d8790a681f60813eb678/vnext/Microsoft.ReactNative/ReactHost/React.h#L89).

windows/ReactTestApp/Manifest.h
tido64: It looks like `at()` may be using `find()` underneath as well so this is fine.

windows/ReactTestApp/Manifest.cpp
```diff
@@ -0,0 +1,45 @@
+#include "pch.h"
+
+#include "Manifest.h"
+
+#include <fstream>
+#include <iostream>
+
+#include <json_custom.h>
```
tido64, line 8: Remove?

windows/ReactTestApp/Manifest.cpp
```diff
@@ -4,8 +4,7 @@
 
 #include <fstream>
 #include <iostream>
-
-#include <json_custom.h>
+#include "nlohmann/json.hpp"
```
tido64, line 7: When `#include`-ing external files, please use `<>` instead.

PR #167, opened 2020-07-13
fix(macOS): `pod install` fails when `react-native` version differs

tido64:
`react-native-macos` does not depend on `react-native` so their version numbers need not be in sync.

The issue manifests itself when you have a project like below:

```
project
├── ios
├── macos
└── node_modules
    ├── react-native@0.62
    └── react-native-macos@0.61
```

In this case, `test_app.rb` will import `use_react_native-0.62.rb` for both iOS and macOS targets. In the latter case, it will fail since 0.61 is installed.

### DIFF
--- a/ios/test_app.rb
+++ b/ios/test_app.rb
@@ -64,6 +64,15 @@ def project_path(file, target_platform)
   File.expand_path(file, File.join(__dir__, '..', target_platform.to_s))
 end
 
+def react_native_path(target_platform)
+  react_native = case target_platform
+                 when :ios then 'react-native'
+                 when :macos then 'react-native-macos'
+                 else raise "Unsupported target platform: #{target_platform}"
+                 end
+  Pathname.new(resolve_module(react_native))
+end
+
 def react_native_pods(version)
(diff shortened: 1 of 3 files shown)


alloy:
Hmm, the change doesn’t really tell me what the problem was and so I can’t really speak to the conceptual fix, but the code itself looks good.

tido64:
> Hmm, the change doesn’t really tell me what the problem was and so I can’t really speak to the conceptual fix, but the code itself looks good.

Sorry, I always forget to write up the full context. The issue manifests itself when you have a project like below:

```
project
├── ios
├── macos
└── node_modules
    ├── react-native@0.62
    └── react-native-macos@0.61
```

In this case, `test_app.rb` will import `use_react_native-0.62.rb` instead of 0.61 for both iOS and macOS targets. In the latter case, it will fail since `react-native-macos` 0.61 is installed. I've updated the PR description and will also include it in the commit message.

alloy:
Ah, gotcha; thanks for the write-up 👍

PR #182, opened 2020-08-05
Initial props parsing

AnastasiaOrishchenko:
nlohmann json library we are using doesn't support parsing to any dynamic types like `std::any` or `folly::dynamic`. That is the reason why parsing of initial properties where value type is unknown has to be written manually.
Furthermore, `InitialProps()` method of `ReactRootView` requires sending initial properties one by one, so type checks need to be done there as well.


windows/ReactTestApp/MainPage.cpp
```diff
@@ -118,4 +131,35 @@ namespace winrt::ReactTestApp::implementation
...
+            }
+            writer.WriteArrayEnd();
+        } else if (propertyValue.type() == typeid(std::map<std::string, std::any>)) {
+            writer.WriteObjectBegin();
+            for (auto &&e : std::any_cast<std::map<std::string, std::any>>(propertyValue)) {
+                writer.WritePropertyName(to_hstring(e.first));
+                WritePropertyValue(e.second, writer);
+            }
+            writer.WriteObjectEnd();
+        }
```
tido64, line 162: Should we add an `else` case here and assert?

windows/ReactTestApp/Manifest.cpp
```diff
@@ -19,6 +21,65 @@ namespace
             return std::nullopt;
         }
     }
+
+    std::any getAny(const nlohmann::json &j)
+    {
+        switch (j.type()) {
+            case value_t::null:
+                return std::nullopt;
+                break;
```
tido64, line 30: Both `return` and `break` are jump instructions, so having both is redundant. Same below...
```suggestion
                return std::nullopt;
```

windows/ReactTestApp/Manifest.cpp
```diff
@@ -19,6 +21,65 @@ namespace
...
+        }
+    }
+
+    std::optional<std::map<std::string, std::any>> parseInitialProps(const nlohmann::json &j)
+    {
+        auto element = j.find("initialProperties");
+        if (element != j.end()) {
+            std::map<std::string, std::any> map;
+            for (auto &&property : element->items()) {
+                map.insert(std::pair(property.key(), getAny(property.value())));
```
tido64, line 75: We should use `std::make_pair` everywhere so the types can be deduced:
```suggestion
                map.insert(std::make_pair(property.key(), getAny(property.value())));
```

windows/ReactTestApp/Manifest.cpp
```diff
@@ -19,6 +21,65 @@ namespace
...
+                for (auto &&e : j.items()) {
+                    array.push_back(getAny(e.value()));
+                }
+                return array;
+                break;
+            }
+            // value_t::discarded. This case should never be hit since the check for malformed json
+            // is done previously in GetManifest()
+            default:
+                return nullptr;
```
tido64, line 65: If this case is not reachable, we should add an assertion (and return a proper value):
```suggestion
            default:
                assert(false);
                return std::nullopt;
```

windows/ReactTestApp/MainPage.cpp
```diff
@@ -45,7 +45,20 @@ namespace winrt::ReactTestApp::implementation
...
+                    ReactRootView().InitialProps(
+                        [firstComponent](winrt::Microsoft::ReactNative::IJSValueWriter writer) {
+                            writer.WriteObjectBegin();
+                            for (auto &&property : firstComponent.initialProperties.value()) {
+                                std::any value = property.second;
+                                writer.WritePropertyName(to_hstring(property.first));
+                                WritePropertyValue(value, writer);
+                            }
+                            writer.WriteObjectEnd();
+                        });
```
tido64, line 60: This only sets initial props for the case where we have a single component. We need to add it for when we switch components as well.

windows/ReactTestApp/MainPage.h
```diff
@@ -31,6 +33,9 @@ namespace winrt::ReactTestApp::implementation
             Windows::Foundation::IInspectable const &);
         void SetUpTitleBar();
     };
+
+    void WritePropertyValue(std::any propertyValue,
+                            winrt::Microsoft::ReactNative::IJSValueWriter writer);
```
tido64, line 38: We should pass in const references to avoid copies:
```suggestion
    void WritePropertyValue(const std::any &propertyValue,
                            const winrt::Microsoft::ReactNative::IJSValueWriter &writer);
```

windows/ReactTestApp/Manifest.cpp
```diff
@@ -7,6 +7,8 @@
 
 #include <nlohmann/json.hpp>
 
+using namespace nlohmann::detail;
```
tido64, line 10: Do we really need the whole namespace? Can we reduce this to:
```suggestion
using nlohmann::json::value_t;
```

windows/ReactTestApp/MainPage.cpp
```diff
@@ -45,7 +45,20 @@ namespace winrt::ReactTestApp::implementation
 
             // If only one component is present load it automatically
             if (components.size() == 1) {
-                ReactRootView().ComponentName(to_hstring(components.at(0).appKey));
+                auto firstComponent = components.at(0);
+                ReactRootView().ComponentName(to_hstring(firstComponent.appKey));
+                if (firstComponent.initialProperties.has_value()) {
+                    ReactRootView().InitialProps(
+                        [firstComponent](winrt::Microsoft::ReactNative::IJSValueWriter writer) {
```
tido64, line 52: Should this be a const reference?
```suggestion
                        [firstComponent](const winrt::Microsoft::ReactNative::IJSValueWriter &writer) {
```

windows/ReactTestApp/MainPage.cpp
```diff
@@ -45,7 +45,20 @@ namespace winrt::ReactTestApp::implementation
...
+                            writer.WriteObjectBegin();
+                            for (auto &&property : firstComponent.initialProperties.value()) {
+                                std::any value = property.second;
```
tido64, line 55: We should avoid a copy here:
```suggestion
                                auto &value = property.second;
```

windows/ReactTestApp/MainPage.cpp
```diff
@@ -100,14 +92,12 @@ namespace winrt::ReactTestApp::implementation
...
-        hstring appKey = to_hstring(component.appKey);
-        ReactTestApp::ComponentViewModel newComponent =
-            winrt::make<ComponentViewModel>(appKey, componentDisplayName);
 
         MenuFlyoutItem newMenuItem;
-        newMenuItem.Text(newComponent.DisplayName());
-        newMenuItem.CommandParameter(newComponent);
-        newMenuItem.Click({this, &MainPage::SetReactComponentName});
+        newMenuItem.Text(componentDisplayName);
+        newMenuItem.Click([this, component](IInspectable const &, RoutedEventArgs) {
```
tido64, line 98: Avoid copying `component`:
```suggestion
        newMenuItem.Click([this, &component](IInspectable const &, RoutedEventArgs) {
```

windows/ReactTestApp/MainPage.cpp
```diff
@@ -131,8 +121,23 @@ namespace winrt::ReactTestApp::implementation
         TitleBar().Height(sender.Height());
     }
 
-    void WritePropertyValue(std::any propertyValue,
-                            winrt::Microsoft::ReactNative::IJSValueWriter writer)
+    void MainPage::WriteInitialProperties(
+        std::optional<std::map<std::string, std::any>> const &initialProps)
+    {
+        if (initialProps.has_value()) {
+            ReactRootView().InitialProps([initialProps](IJSValueWriter const &writer) {
```
tido64, line 128: Avoid copying `initialProps`:
```suggestion
            ReactRootView().InitialProps([&initialProps](IJSValueWriter const &writer) {
```

windows/ReactTestApp/MainPage.cpp
```diff
@@ -73,10 +75,13 @@ namespace winrt::ReactTestApp::implementation
...
     }
 
-    void MainPage::SetReactComponentName(IInspectable const &sender, RoutedEventArgs)
+    void
+    MainPage::LoadReactComponent(std::string const &appKey,
+                                 std::optional<std::map<std::string, std::any>> const &initialProps)
     {
-        auto s = sender.as<MenuFlyoutItem>().CommandParameter();
-        ReactRootView().ComponentName(s.as<ComponentViewModel>()->AppKey());
+        ReactRootView().ReactNativeHost().ReloadInstance();
```
tido64, line 82: Why do we have to reload the instance?

windows/ReactTestApp/MainPage.cpp
```diff
@@ -87,14 +92,12 @@ namespace winrt::ReactTestApp::implementation
     MenuFlyoutItem MainPage::MakeComponentMenuButton(::ReactTestApp::Component &component)
```
tido64, line 92: Probably should've noticed this earlier but can `component` be `const`?
```suggestion
    MenuFlyoutItem MainPage::MakeComponentMenuButton(::ReactTestApp::Component const &component)
```

windows/ReactTestApp/MainPage.h
```diff
@@ -22,15 +24,20 @@ namespace winrt::ReactTestApp::implementation
     private:
         ::ReactTestApp::ReactInstance reactInstance_;
 
-        void SetReactComponentName(Windows::Foundation::IInspectable const &,
-                                   Windows::UI::Xaml::RoutedEventArgs);
+        void LoadReactComponent(std::string const &appKey,
+                                std::optional<std::map<std::string, std::any>> const &initialProps);
+        void
+        WriteInitialProperties(std::optional<std::map<std::string, std::any>> const &initialProps);
```
tido64, line 30: Should this be `Set*` instead?
```suggestion
        void
        SetInitialProperties(std::optional<std::map<std::string, std::any>> const &initialProps);
```

windows/ReactTestApp/MainPage.cpp
```diff
@@ -118,4 +121,52 @@ namespace winrt::ReactTestApp::implementation
...
+            writer.WriteBoolean(std::any_cast<boolean>(propertyValue));
+        } else if (propertyValue.type() == typeid(std::int64_t)) {
+            writer.WriteInt64(std::any_cast<std::int64_t>(propertyValue));
+        } else if (propertyValue.type() == typeid(std::uint64_t)) {
+            writer.WriteInt64(std::any_cast<std::uint64_t>(propertyValue));
+        } else if (propertyValue.type() == typeid(double)) {
+            writer.WriteDouble(std::any_cast<double>(propertyValue));
+        } else if (propertyValue.type() == typeid(std::nullopt)) {
+            writer.WriteNull();
+        } else if (propertyValue.type() == typeid(std::basic_string<char>)) {
```
tido64, line 152: `std::string` is the same as `std::basic_string<char>` but is more legible:
```suggestion
        } else if (propertyValue.type() == typeid(std::string)) {
```

windows/ReactTestApp/MainPage.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <any>
```
tido64, line 3: We should always `#include` what we use:
```suggestion
#include <any>
#include <optional>
#include <map>
#include <string>
```

windows/ReactTestApp/MainPage.cpp
```diff
@@ -43,12 +43,14 @@ namespace winrt::ReactTestApp::implementation
...
             }
 
+            ReactRootView().ReactNativeHost(reactInstance_.ReactHost());
+
             // If only one component is present load it automatically
             if (components.size() == 1) {
-                ReactRootView().ComponentName(to_hstring(components.at(0).appKey));
+                auto firstComponent = components.at(0);
+                ReactRootView().ComponentName(to_hstring(firstComponent.appKey));
+                WriteInitialProperties(firstComponent.initialProperties);
```
tido64, line 52: If we can move the `ReloadInstance()` call to `MenuFlyoutItem` click handler, we can reuse `LoadReactComponent` here:
```suggestion
                auto firstComponent = components.at(0);
                LoadReactComponent(firstComponent.appKey, firstComponent.initialProperties);
```

windows/ReactTestApp/MainPage.cpp
AnastasiaOrishchenko: When passing `component` by reference, component.appKey is empty string and component.initialProperties is `nullopt` by the time the code in click handler runs. Could it be that component object is destructed?

tido64: Did we make sure to pass `component` by reference all the way? If we made a copy somewhere, this would fail.

tido64: Oh, yes... We should make a copy here since `manifest` will be destructed when `MainPage::MainPage()` returns:
```cpp
    MainPage::MainPage()
    {
        ...
        std::optional<::ReactTestApp::Manifest> manifest = ::ReactTestApp::GetManifest();
        ...
    }
```

windows/ReactTestApp/MainPage.cpp
```diff
@@ -118,4 +120,52 @@ namespace winrt::ReactTestApp::implementation
         TitleBar().Height(sender.Height());
     }
 
+    void MainPage::SetInitialProperties(
+        std::optional<std::map<std::string, std::any>> const &initialProps)
+    {
+        if (initialProps.has_value()) {
+            ReactRootView().InitialProps([&initialProps](IJSValueWriter const &writer) {
```
tido64, line 127: We should always set `InitialProps` even if it has no value otherwise we won't be able to load the same component with different props since `ReactRootView` won't trigger a reload for the same component name:
```json
{
  "name": "Example",
  "displayName": "Example",
  "components": [
    {
      "appKey": "Example",
      "displayName": "App with no props"
    },
    {
      "appKey": "Example",
      "displayName": "App with some props",
      "initialProperties": {
        "something": "nothing"
      }
    }
  ]
}
```

PR #187, opened 2020-08-10
Bootstrapping Instrumented tests for Android

arazabishov:
* Made necessary changes to RNTA on Android to be able to run it without a test component
* Configured the project to run RNTA tests only when executed from the RNTA project to avoid consumers running our tests
* Setting-up a job to run instrumented tests on CI
* Added one instrumented test for Manifest parsing (TODO: add more to match iOS)


android/app/build.gradle
```diff
@@ -36,10 +36,13 @@ def testAppDir = file("$projectDir/../../")
...
 applyTestAppModule(project, "com.react.testapp")
 
+project.ext.isRntaProject = projectDir.getParent() != null &&
+    (rootDir == file(projectDir.getParent()))
+
 project.ext.react = [
     applicationId: getApplicationId(),
     enableFlipper: getFlipperVersion(rootDir),
-    enableHermes: true,
+    enableHermes : true,
```
tido64, line 45: Revert?

android/app/build.gradle
```diff
@@ -74,18 +77,26 @@ android {
...
         }
     }
+
+    testOptions {
+        unitTests.all {
+            if (!project.ext.isRntaProject) {
+                exclude "**/com/react/testapp/**"
+            }
+        }
+    }
```
tido64, line 87: Would it be cleaner to exclude the tests in `.npmignore`?

android/app/build.gradle
```diff
@@ -113,13 +124,13 @@ dependencies {
...
-            exclude group:'com.facebook.fbjni'
+            exclude group: 'com.facebook.fbjni'
         }
         debugImplementation("com.facebook.flipper:flipper-fresco-plugin:${flipperVersion}") {
-            exclude group:'com.facebook.flipper'
+            exclude group: 'com.facebook.flipper'
         }
         debugImplementation("com.facebook.flipper:flipper-network-plugin:${flipperVersion}") {
-            exclude group:'com.facebook.flipper'
+            exclude group: 'com.facebook.flipper'
```
tido64, line 133: Revert?

.github/workflows/build.yml
```diff
@@ -164,6 +164,24 @@ jobs:
...
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8    
+    - name: Run instrumented tests
+      uses: reactivecircus/android-emulator-runner@v2
+      with:        
+        api-level: ${{ matrix.api-level }}
+        script: cd android/ && sh gradlew --no-daemon clean build connectedCheck --info
+        arch: x86_64
```
tido64, line 184: Can we run this as part of the `android` job? What's the difference between all the API levels and do we really need to test them all?

test-app.gradle
```diff
@@ -34,6 +35,11 @@ ext.applyTestAppModule = { Project project, String packageName ->
     applyNativeModulesAppBuildGradle(project, packageName)
 
     def manifestFile = findFile("app.json")
+    if (manifestFile == null) {
+        logger.info("Manifest was not found; test app integration is disabled")
+        return
+    }
```
tido64, line 41: Is this here so we can run tests?

android/app/build.gradle
arazabishov: This is the result of running the AS formatter. It aligns columns using the `:` char as delimeter.

android/app/build.gradle
arazabishov: This doesn't exclude tests from `node_modules`, it is a flag to exclude a group of tests within the source directory. Tests are running directly against `rnta/android`, not against `example/node_modules/rnta/android`.

android/app/build.gradle
arazabishov: See the comment above.

.github/workflows/build.yml
arazabishov: We should test 21 and 29 at least. It can't be a part of android as it runs against both mac and windows (hardware acceleration is available only on mac).

test-app.gradle
arazabishov: That too, but also to be able to dev `rnta/android` without having a test component consuming it.

android/app/build.gradle
tido64: Not quite sure I follow. If the tests are excluded from the published packages, do we still need this block here? Within RNTA, we'll still have access to them as intended.

test-app.gradle
tido64: Can we change "Manifest" to "app.json"? And should this be more strongly worded and be a warning at the very least? As this is a RNTA only use case, it would be really confusing to consumers. Is there some way we can only warn if it's not being run by RNTA contributors and the CI?

.github/workflows/build.yml
arazabishov: We can leave only 29 if CI times are more important rather than possibility of breaking something on v21 (which is not very likely taking into account the subset of android APIs that we're using).

.github/workflows/build.yml
```diff
@@ -164,6 +164,37 @@ jobs:
...
+      uses: actions/setup-node@v1
+      with:
+        # Node has a bug where it crashes compiling a regular
+        # expression of react-native cli. Using a specific
+        # Node version helps to workaround this problem:
+        # https://github.com/facebook/react-native/issues/26598
+        node-version: 12.9.1
+    - uses: actions/checkout@v2
+    - name: Validate Gradle wrapper
+      uses: gradle/wrapper-validation-action@v1
```
tido64, line 189: Should we skip this step?

.github/workflows/build.yml
```diff
@@ -164,6 +164,37 @@ jobs:
...
         working-directory: example
+  android-test:
+    name: "Android: instrumented tests"
+    strategy:
+      matrix:
+        api-level: [21, 29]
+    runs-on: macos-latest
+    if: "!contains(github.event.head_commit.message, '[skip ci]')"
+    steps:
+    - name: Set up JDK 1.8
```
tido64, line 175: I think we need to indent this block (or just run Prettier so it can format this file):
```suggestion
      - name: Set up JDK
```

.github/workflows/build.yml
```diff
@@ -164,6 +164,37 @@ jobs:
...
+        # Node has a bug where it crashes compiling a regular
+        # expression of react-native cli. Using a specific
+        # Node version helps to workaround this problem:
+        # https://github.com/facebook/react-native/issues/26598
+        node-version: 12.9.1
+    - uses: actions/checkout@v2
+    - name: Validate Gradle wrapper
+      uses: gradle/wrapper-validation-action@v1
+    - name: Install
+      run: yarn
```
tido64, line 191: We should use the node_modules cache:
```suggestion
      - name: Cache /node_modules
        uses: actions/cache@v2
        with:
          path: node_modules
          key: project-${{ runner.os }}-${{ hashFiles('yarn.lock') }}
      - name: Install
        run: yarn
```

android/app/build.gradle
arazabishov: Oh, now I see! Yes, the approach with `.npmignore` will be better. I will take a look.

test-app.gradle
arazabishov: Updated the message and switched to the `warning` logging level. I think it is possible to detect if we're running under RNTA project. Would you prefer to fail hard if `app.json` is missing for consumers?

.github/workflows/build.yml
```diff
@@ -164,6 +164,22 @@ jobs:
           pushd android 1> /dev/null
           ./gradlew clean build check test
         working-directory: example
+      - name: Cache node_modules
```
arazabishov, line 167: I didn't find other syntax to in GH actions to avoid duping `if` statements :(

.github/workflows/build.yml
```diff
@@ -164,6 +164,22 @@ jobs:
...
+        if: ${{ matrix.os == 'macos-latest' }}
+        uses: actions/cache@v2
+        with:
+          path: node_modules
+          key: rnta-${{ runner.os }}-${{ hashFiles('yarn.lock') }}
```
tido64, line 172: We should use the same key as elsewhere so we can reuse the same cache:
```suggestion
          key: project-${{ runner.os }}-${{ hashFiles('yarn.lock') }}
```

test-app.gradle
tido64: Did it fail previously? I think we want to keep that behaviour.

.github/workflows/build.yml
tido64: That's fine. I forgive you 😛

.github/workflows/build.yml
arazabishov: Do we? In other cases we were caching `example/node_modules` based on the `example/yarn.lock` file. Here I am caching the `node_modules` folder from the project root.

tido64: Yes, we are caching root `node_modules` folder in
https://github.com/microsoft/react-native-test-app/blob/5d75836ccc116d425166b904b56591f1ee1015c7/.github/workflows/build.yml#L21
https://github.com/microsoft/react-native-test-app/blob/5d75836ccc116d425166b904b56591f1ee1015c7/.github/workflows/build.yml#L100
https://github.com/microsoft/react-native-test-app/blob/5d75836ccc116d425166b904b56591f1ee1015c7/.github/workflows/build.yml#L206
https://github.com/microsoft/react-native-test-app/blob/5d75836ccc116d425166b904b56591f1ee1015c7/.github/workflows/build.yml#L274
https://github.com/microsoft/react-native-test-app/blob/5d75836ccc116d425166b904b56591f1ee1015c7/.github/workflows/build.yml#L320

android/app/src/androidTest/java/com/react/testapp/ManifestTest.kt
```diff
@@ -0,0 +1,69 @@
...
+        adapter = ManifestJsonAdapter(
+            Moshi.Builder()
+                .add(MoshiBundleAdapter())
+                .build()
+        )
+    }
+
+    @Test
+    fun parsesMultipleComponentsWithProps() {
+        val json = """
```
tido64, line 27: Can we add an issue to consolidate all the manifest parsing tests?

arazabishov: Do you refer to what we have discussed during the last sync?

tido64: Yes

arazabishov: https://github.com/microsoft/react-native-test-app/issues/193

tido64: Thanks ❤️

PR #192, opened 2020-08-12
Consuming as npm package

AnastasiaOrishchenko:
Added script to create windows test app project and solution dynamically. 
Running following command from `example` folder will resolve paths to installation place of react-native-windows node module and all required assets and create the solution.
`node ..\windows\test-app.js`


windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+import * as fs from 'fs';
+
+const windowsDir = 'windows';
+const reactNativeModulePath = findClosestPathTo(path.join('node_modules', 'react-native-windows'));
+const testAppNodeModulePath = findClosestPathTo(path.join('node_modules', 'react-native-test-app'));
+const srcRootPath = path.join(testAppNodeModulePath, windowsDir);
+const destPath = path.resolve('');
+
+
+function findClosestPathTo(fileOrDirName: string): string {
```
tido64, line 11: Since you're returning `null`, the return type should be:
```suggestion
function findClosestPathTo(fileOrDirName: string): string | null {
```

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+
+    return null
+}
+
+/**
+ * Text to replace, + config options
+ */
+export type Replacements = {
+    regExpPatternsToRemove?: RegExp[];
+    [key: string]: any;
```
tido64, line 32: We should avoid using `any` when we know the type:
```suggestion
    [key: string]: string;
```

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+        }
+        //Get parent folder
+        basePath = path.dirname(basePath)
+    }
+
+    return null
+}
+
+/**
+ * Text to replace, + config options
```
tido64, line 28: Redundant comma:
```suggestion
 * Text to replace + config options
```

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
+import * as path from 'path';
```
tido64, line 1: Can we add tests?

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+
+    Object.keys(replacements).forEach(regex => {
+        content = content.replace(new RegExp(regex, 'g'), replacements[regex]);
+    });
+
+    return content;
+}
+
+// Binary files, don't process these (avoid decoding as utf8)
+const binaryExtensions = ['.png', '.jar', '.keystore'];
```
tido64, line 67: This doesn't look very extensive. Is this list based on the files we have in the repo? If so, why do we have `.jar` here?

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+            encoding: 'utf8',
+            mode: srcPermissions,
+        });
+    }
+}
+
+/**
+ * Same as 'cp' on Unix. Don't do any replacements.
+ */
+function copyBinaryFile(
```
tido64, line 112: Wouldn't it be simpler to use [`fs.copyFile`](https://nodejs.org/api/fs.html#fs_fs_copyfile_src_dest_mode_callback)?

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+        const filename = path.relative(srcPath, absoluteSrcFilePath);
+        const relativeDestPath = path.join(relativeDestDir, filename);
+        await copyAndReplace(
+            absoluteSrcFilePath,
+            destPath,
+            relativeDestPath,
+            replacements
+        );
+    }
+}
```
tido64, line 163: We should allow `copyAndReplace` to be run in parallel or fire-and-forgot:
```suggestion
export function copyAndReplaceAll(
    srcPath: string,
    destPath: string,
    relativeDestDir: string,
    replacements: Replacements = {}
): Promise<void> {
    return Promise.all(walk(srcPath).map((absoluteSrcFilePath) => {
        const filename = path.relative(srcPath, absoluteSrcFilePath);
        const relativeDestPath = path.join(relativeDestDir, filename);
        return copyAndReplace(
            absoluteSrcFilePath,
            destPath,
            relativeDestPath,
            replacements
        );
    }));
}
```

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+        throw new Error('Need a path to copy to');
+    }
+
+    if (!reactNativeModulePath) {
+        throw new Error('react-native-windows node module is not installed');
+    }
+
+    if (!testAppNodeModulePath) {
+        throw new Error('react-native-test-app node module is not installed');
+    }
```
tido64, line 201: These should all be parameters to this function.

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+
+    const projDir = 'ReactTestApp';
+
+    createDir(path.join(destPath, windowsDir));
+    createDir(path.join(destPath, windowsDir, projDir));
+    createDir(path.join(destPath, windowsDir, projDir, 'Bundle'));
```
tido64, line 207: We should use [`fs.mkdirSync`](https://nodejs.org/api/fs.html#fs_fs_mkdirsync_path_options) here:
```suggestion
    fs.mkdirSync(path.join(destPath, windowsDir, projDir, 'Bundle'), { 'recursive': true });
```

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+    //Read path to resources from manifest and copy them
+    fs.readFile(manifestFilePath, async function (error, content) {
+        var resourcesPaths = JSON.parse(content.toString()).resources;
+        if (!resourcesPaths) {
+            throw new Error('No resources specified in app.json');
+        }
+        resourcesPaths = resourcesPaths.windows ? resourcesPaths.windows : resourcesPaths;
+        for (const resource of resourcesPaths) {
+            await copyAndReplaceAll(path.relative(path.dirname(manifestFilePath), resource), destPath, path.join(windowsDir, projDir, 'Bundle', path.basename(resource)));
+        }
```
tido64, line 220:
```suggestion
        await Promise.all(resourcesPaths.map((resource) => {
            return copyAndReplaceAll(
                path.relative(path.dirname(manifestFilePath), resource),
                destPath,
                path.join(windowsDir, projDir, 'Bundle', path.basename(resource))
            );
        }));
```

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+    createDir(path.join(destPath, windowsDir, projDir, 'Bundle'));
+
+    const manifestFilePath = findClosestPathTo('app.json');
+
+    //Read path to resources from manifest and copy them
+    fs.readFile(manifestFilePath, async function (error, content) {
+        var resourcesPaths = JSON.parse(content.toString()).resources;
+        if (!resourcesPaths) {
+            throw new Error('No resources specified in app.json');
+        }
```
tido64, line 216: I think we should allow empty `resources`.

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+        resourcesPaths = resourcesPaths.windows ? resourcesPaths.windows : resourcesPaths;
+        for (const resource of resourcesPaths) {
+            await copyAndReplaceAll(path.relative(path.dirname(manifestFilePath), resource), destPath, path.join(windowsDir, projDir, 'Bundle', path.basename(resource)));
+        }
+    });
+
+    const projectFilesReplacements: Record<string, any> = {
+        regExpPatternsToRemove: ['//\\sclang-format\\s(on|off)\\s'],
```
tido64, line 224: Why do we remove this?

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+                windowsDir,
+                projDir,
+                'ReactTestApp_TemporaryKey.pfx',
+            ),
+        },
+        {
+            from: path.join(srcRootPath, projDir, 'packages.config'),
+            to: path.join(windowsDir, projDir, 'packages.config'),
+        },
+    ];
```
tido64, line 264: Can we generalize it like below?
```suggestion
    const projectFilesMappings = [
        'Package.appxmanifest',
        'packages.config',
        'PropertySheet.props',
        'ReactTestApp_TemporaryKey.pfx',
        'ReactTestApp.vcxproj.filters',
        'ReactTestApp.vcxproj'
    ].map((file) => ({
        from: path.join(srcRootPath, projDir, file),
        to: path.join(windowsDir, projDir, file),
    }));
```

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+
+    for (const mapping of projectFilesMappings) {
+        await copyAndReplace(
+            mapping.from,
+            destPath,
+            mapping.to,
+            projectFilesReplacements
+        );
+    }
```
tido64, line 273: Do we really have to wait for these to finish before we can move on? Otherwise, I think we should fire and forget:
```suggestion
    projectFilesMappings.forEach(({ from, to }) => {
        copyAndReplace(from, destPath, to, projectFilesReplacements);
    });
```
You can also destructure `mapping` in the argument list to save some characters 😄

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+        path.join(srcRootPath, projDir, 'Assets'),
+        destPath,
+        path.join(windowsDir, projDir, 'Assets')
+    );
+
+    await copySourceFiles(
+        path.join(srcRootPath, projDir),
+        destPath,
+        path.join(windowsDir, projDir)
+    );
```
tido64, line 296: Fire and forget?

tido64, line 296: Do we have to copy them? Can we simply reference them from `.vcxproj`?

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+    destPath: string,
+    relativeDestPath: string,
+    replacements: Replacements = {}
+) {
+    const fullDestPath = path.join(destPath, relativeDestPath);
+    if (fs.lstatSync(srcPath).isDirectory()) {
+        if (!fs.existsSync(fullDestPath)) {
+            fs.mkdirSync(fullDestPath);
+        }
+        return;
```
tido64, line 86: Why do we return here?

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+    srcPath: string,
+    replacements: Replacements,
+): string {
+    let content = fs.readFileSync(srcPath, 'utf8');
+
+    Object.keys(replacements).forEach(regex => {
+        content = content.replace(new RegExp(regex, 'g'), replacements[regex]);
+    });
+
+    return content;
```
tido64, line 63: This can be simplified further by using `reduce`:
```suggestion
    const content = fs.readFileSync(srcPath, 'utf8');
    return Object.keys(replacements).reduce((regex) => {
        return content.replace(new RegExp(regex, 'g'), replacements[regex]);
    }, content);
```

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+const windowsDir = 'windows';
+const reactNativeModulePath = findClosestPathTo(path.join('node_modules', 'react-native-windows'));
+const testAppNodeModulePath = findClosestPathTo(path.join('node_modules', 'react-native-test-app'));
+const srcRootPath = path.join(testAppNodeModulePath, windowsDir);
+const destPath = path.resolve('');
+
+
+function findClosestPathTo(fileOrDirName: string): string {
+    var basePath = path.resolve('');
+    var rootDirectory = basePath.split(path.sep)[0] + path.sep;
```
tido64, line 13: Prefer to use `const` or `let`. `var`s aren't properly scoped.
```suggestion
    let basePath = path.resolve('');
    const rootDirectory = basePath.split(path.sep)[0] + path.sep;
```

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+function walk(current: string): string[] {
+    if (!fs.lstatSync(current).isDirectory()) {
+        return [current];
+    }
+
+    const files = fs
+        .readdirSync(current)
+        .map(child => walk(path.join(current, child)));
+    const result: string[] = [];
+    return result.concat.apply([current], files);
```
tido64, line 44: We can use the spread operator to make this more legible:
```suggestion
    return [current, ...files];
```

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+            if (err) {
+                throw err;
+            }
+        });
+    } else {
+        // Text file
+        const srcPermissions = fs.statSync(srcPath).mode;
+        let content = resolveContents(srcPath, replacements);
+
+        fs.writeFileSync(fullDestPath, content, {
```
tido64, line 102: Do we really need to wait for this write to finish?
```suggestion
        fs.writeFile(fullDestPath, content, {
```

windows/test-app.ts
```diff
@@ -0,0 +1,300 @@
...
+}
+
+const cppSourceFilesExtensions = ['.h', '.cpp', '.idl', '.xaml'];
+
+export async function copySourceFiles(
+    srcPath: string,
+    destPath: string,
+    relativeDestDir: string
+) {
+    for (const absoluteSrcFilePath of fs.readdirSync(srcPath).map(_ => path.join(srcPath, _))) {
```
tido64, line 172: Same here: We should allow `copyAndReplace` to be run in parallel or fire-and-forgot.

windows/ReactTestApp/ReactTestApp.vcxproj
```diff
@@ -70,19 +68,24 @@
         <WholeProgramOptimization>true</WholeProgramOptimization>
         <LinkIncremental>false</LinkIncremental>
     </PropertyGroup>
+    <PropertyGroup Label="UserMacros">
+        <BundleContentRoot>..\..\example\dist</BundleContentRoot>
+        <ManifestRootPath>..\..\example\</ManifestRootPath>
+        <BundleContent>$(BundleContentRoot)\**\*</BundleContent>
+        <ReactNativeModulePath>..\..\node_modules\react-native-windows\</ReactNativeModulePath>
+    </PropertyGroup>
```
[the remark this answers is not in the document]
AnastasiaOrishchenko: This is needed if we want to run app directly from react-native-test-app/windows. If we will only run generated application, then this can be removed. Furthermore, we can then simplify solution file as well, so they will become more like templates for generating apps.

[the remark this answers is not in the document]
tido64: For all the other platforms, we are developing them through the Example app. If we are able to do that now, I would remove this block and simplify things as you've said.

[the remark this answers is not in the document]
AnastasiaOrishchenko: I think it should be possible if we reference source files, like you said [here](https://github.com/microsoft/react-native-test-app/pull/192#discussion_r469099065).

windows/test-app.mjs
```diff
@@ -0,0 +1,228 @@
+//@ts-check
```
AnastasiaOrishchenko, line 1: Loading module is experimental feature in Node 12, so running this script will give a warning. Any suggestion on what we should do?

windows/test-app.mjs
```diff
@@ -0,0 +1,228 @@
...
+            destPath,
+            relativeDestPath,
+            replacements
+        );
+    }));
+}
+
+const cppSourceFilesExtensions = ['.h', '.cpp', '.idl', '.xaml'];
+
+export function linkSourceFiles(
```
AnastasiaOrishchenko, line 123: Linking source files, so we can continue developing from example app.

windows/test-app.mjs
tido64: You can use `require()` instead:
```js
const path = require('path');
const fs = require('fs');
```

windows/test-app.mjs
```diff
@@ -0,0 +1,228 @@
...
+
+function walk(current) {
+    if (!fs.lstatSync(current).isDirectory()) {
+        return [current];
+    }
+
+    const files = fs
+        .readdirSync(current)
+        .map(child => walk(path.join(current, child)));
+    const result = [];
```
tido64, line 36: This seems unused. We can remote it.

windows/test-app.mjs
tido64: Do we have to create links? Can't we reference them from `.vcxproj`?

windows/test-app.mjs
```diff
@@ -0,0 +1,228 @@
...
+        'ReactTestApp.vcxproj.filters',
+        'ReactTestApp_TemporaryKey.pfx',
+        'packages.config'
+    ].map((file) => ({
+        from: path.join(srcRootPath, projDir, file),
+        to: path.join(windowsDir, projDir, file)
+    }));
+
+    for (const mapping of projectFilesMappings) {
+        await copyAndReplace(
```
tido64, line 199: Do we have to wait for the copy/replace to complete? If not, we can remove `await`.

windows/test-app.mjs
tido64, line 1: Can you run `yarn prettier --write windows/test-app.mjs` so we can get this file formatted?

tido64: Also, why the `.mjs` extension? Is that something new? Can we use `.js` instead?

windows/test-app.mjs
AnastasiaOrishchenko: We don't need to create links if it's convenient to edit source files from `react-native-test-app/windows` folder while developing. If we want to have both project files and source files in the same folder for development purposes, then we should create links.

windows/test-app.mjs
AnastasiaOrishchenko: Using `.js` gives warning ```
To load an ES module, set "type": "module" in the package.json or use the
.mjs extension.
```

Comment (tido64) [reply windows/test-app.mjs:1]: I think if you replace `import` with `require`, it should go away.

File: windows/test-app.js [hunk 27]
```diff
@@ -0,0 +1,258 @@
...
+        // Binary file
+        fs.copyFile(srcPath, fullDestPath, (err) => {
+            if (err) {
+                throw err;
+            }
+        });
+    } else {
+        // Text file
+        const srcPermissions = fs.statSync(srcPath).mode;
+        let content = resolveContents(srcPath, replacements);
```
tido64, line 90: If we don't plan to replace `content`, we should make it `const`:
```suggestion
        const content = resolveContents(srcPath, replacements);
```

windows/test-app.js
```diff
@@ -0,0 +1,258 @@
...
+    const projDir = "ReactTestApp";
+
+    fs.mkdirSync(path.join(destPath, windowsDir, projDir, "Bundle"), {
+        recursive: true,
+    });
+
+    const manifestFilePath = findClosestPathTo("app.json");
+
+    //Read path to resources from manifest and copy them
+    fs.readFile(manifestFilePath, function (error, content) {
```
tido64, line 183: Use lambda notation:
```suggestion
    fs.readFile(manifestFilePath, (error, content) => {
```

windows/test-app.js
```diff
@@ -0,0 +1,258 @@
...
+        var resourcesPaths = JSON.parse(content.toString()).resources;
```
tido64, line 184: Prefer `let` over `var`:
```suggestion
        let resourcesPaths = JSON.parse(content.toString()).resources;
```

windows/test-app.js
```diff
@@ -0,0 +1,258 @@
...
+        if (resourcesPaths && !Array.isArray(resourcesPaths)) {
+            resourcesPaths = resourcesPaths.windows ? resourcesPaths.windows : [];
+        } else if (!resourcesPaths) {
+            resourcesPaths = [];
+        }
```
tido64, line 189: I think we can simplify this further:
```suggestion
        resourcesPaths = resourcesPaths.windows || resourcesPaths;
        if (!Array.isArray(resourcesPaths)) {
          return;
        }
```

windows/test-app.js
```diff
@@ -0,0 +1,258 @@
...
+                            throw err;
+                        }
+                    });
+                }
+            }
+        })
+    );
+}
+
+async function copyProjectTemplateAndReplace(
```
tido64, line 152: We don't call `await` here, so we don't need `async`:
```suggestion
function copyProjectTemplateAndReplace(
```

windows/test-app.js
```diff
@@ -0,0 +1,258 @@
+//@ts-check
```
tido64, line 1: Can we also add tests for this file?

windows/test-app.js
```diff
@@ -0,0 +1,258 @@
...
+    destPath,
+    relativeDestPath,
+    replacements = {}
+) {
+    const fullDestPath = path.join(destPath, relativeDestPath);
+    if (fs.lstatSync(srcPath).isDirectory()) {
+        if (!fs.existsSync(fullDestPath)) {
+            fs.mkdirSync(fullDestPath);
+        }
+        return;
```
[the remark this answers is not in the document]
AnastasiaOrishchenko: This method is first called for the folder itself and then for each file/folder inside it from `copyAndReplaceAll`. This is because of the way `walk` method recursively creates a list of all items at given path including both folders and their contents.

windows/test-app.js
AnastasiaOrishchenko: Wouldn't `resourcesPaths.windows || resourcesPaths` fail when key "resources" is absent in manifest file? `resourcesPaths && resourcesPaths.windows || resourcesPaths` should fix it though.

windows/test-app.js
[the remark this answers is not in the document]
tido64: I see, I completely missed that. Thanks 👍

tido64:
I'm getting this error running the script on a clean checkout:
```
$ node.exe node_modules/react-native-test-app/windows/test-app.js
internal/fs/utils.js:269
    throw err;
    ^

Error: ENOENT: no such file or directory, lstat 'dist\assets'
    at Object.lstatSync (fs.js:1007:3)
    at walk (D:\Source\react-native-test-app\example\node_modules\react-native-test-app\windows\test-app.js:32:11)
    at copyAndReplaceAll (D:\Source\react-native-test-app\example\node_modules\react-native-test-app\windows\test-app.js:114:5)
    at D:\Source\react-native-test-app\example\node_modules\react-native-test-app\windows\test-app.js:191:7
    at FSReqCallback.readFileAfterClose [as oncomplete] (internal/fs/read_file_context.js:63:3) {
  errno: -4058,
  syscall: 'lstat',
  code: 'ENOENT',
  path: 'dist\\assets'
}
```
On other platforms, we skip files that don't exist. This is so that users can go straight to `yarn start:windows` without having to bundle first.

windows/test-app.js
```diff
@@ -190,21 +165,22 @@ function copyProjectTemplateAndReplace(
     for (const resource of resourcesPaths) {
       copyAndReplaceAll(
         path.relative(path.dirname(manifestFilePath), resource),
-        destPath,
-        path.join(windowsDir, projDir, "Bundle", path.basename(resource))
+        projectFilesDestPath,
+          path.join(bundleDir, path.basename(resource))
       );
```
tido64: Can we avoid copying assets to `node_modules/.generated/windows/ReactTestApp/Bundle`, and instead create reference to them in `.vcxproj`? I think it will be confusing to users if they have to run `test-app.js` every time they update any assets.

windows/test-app.js
```diff
@@ -241,23 +219,11 @@ function copyProjectTemplateAndReplace(
...
-    destPath,
-    path.join(windowsDir, projDir)
-  );
 }
 
 copyProjectTemplateAndReplace(
-  srcRootPath,
-  destPath,
+    destPath,
+    closestNodeModules,
```
tido64: It looks like your editor is messing with the formatting again. You should run `yarn prettier` to have this file re-formatted.

AnastasiaOrishchenko: Forgot to run it before pushing. Should be fine now.

windows/test-app.js
AnastasiaOrishchenko: Sure, we could do that. Users would still need to rerun the script each time they change resources paths in manifest, but that will probably not happen so frequently.

tido64: Yeah, we can only add paths to resources we know of. 😄

windows/ReactTestApp/ReactTestApp.vcxproj
```diff
@@ -116,104 +141,113 @@
...
+            <DeploymentContent>true</DeploymentContent>
+        </None>
+    </ItemGroup>
+    <ItemGroup>
+        <BundleFileContentToBePackaged Include="$(BundleFileAssets)"/>
+        <None Include="@(BundleFileContentToBePackaged)">
             <Link>$([MSBuild]::MakeRelative($(ProjectDir),'Bundle\%(RecursiveDir)%(Filename)%(Extension)'))</Link>
             <DeploymentContent>true</DeploymentContent>
         </None>
     </ItemGroup>
```
AnastasiaOrishchenko, line 218: This can probably be written in a less cumbersome way. What we try to achieve here is the following:

- When resource entry in manifest is a folder, we want to include the folder itself, not only its contents. For example, for entry `dist/assets` we want the paths to included files to be `Bundle/assets/...`, not just `Bundle/...`
- When entry i a file, we want to include just the file itself (without putting it in an additional directory). So for entry `dist/main.bundle`, the path to included file should be `Bundle/main.bundle`, not `Bundle/dist/main.bundle`

However, I'm not sure how to generalize this.

windows/test-app.js
```diff
@@ -0,0 +1,234 @@
...
+      const resourceSrcPath = path.relative(
+        path.dirname(manifestFilePath),
+        resource
+      );
+      if (fs.existsSync(resourceSrcPath)) {
+        let relativeResourcePath = path.relative(
+          projectFilesDestPath,
+          resourceSrcPath
+        );
+        if (path.extname(resourceSrcPath) === "") {
```
tido64, line 153: Not all files have an extension. We should use `isDirectory()` instead.

windows/test-app.js
```diff
@@ -0,0 +1,234 @@
...
+  );
+
+  fs.mkdirSync(projectFilesDestPath, { recursive: true });
+  fs.mkdirSync(path.join(destPath, windowsDir), { recursive: true });
+
+  const manifestFilePath = findClosestPathTo("app.json");
+
+  let bundleDirContent = "";
+  let bundleFileContent = "";
+  //Read path to resources from manifest and copy them
```
tido64, line 133: We don't really copy them any longer 😛

windows/test-app.js
```diff
@@ -0,0 +1,234 @@
...
+          bundleDirContent = bundleDirContent.concat(
+            relativeResourcePath + ";"
+          );
+        } else {
+          bundleFileContent = bundleFileContent.concat(
+            relativeResourcePath + ";"
+          );
+        }
+      } else {
+        console.log(
```
tido64, line 164: We should use `console.warn()`:
```suggestion
        console.warn(
```

windows/ReactTestApp/ReactTestApp.vcxproj
tido64: Normally, we just want to reference the last "part" of the path. E.g. for `dist/assets`, what we really want to be part of the bundle is `assets`. For a longer path, say `dist/resources/myAsset.res`, we should only reference `myAsset.res`. The final bundle would look something like:
```
Bundle
├── assets
│   └── <content of assets>
├── main.windows.bundle
└── myAsset.res
```
We're keeping only the folder structure inside the referenced folder. Otherwise, we will always flatten everything.

AnastasiaOrishchenko: That's what happens now as well. The problem is that to include all contents of `assets` we need to have `assets\**\*` as include path, which would then match each path inside `assets`. So the last "part" will actually be filename.

tido64 (approved):

PR #194, opened 2020-08-13
Add debug menu option and key accelerators to React component items

AnastasiaOrishchenko:


windows/ReactTestApp/MainPage.cpp
```diff
@@ -82,13 +88,37 @@ namespace winrt::ReactTestApp::implementation
                 AppTitle().Text(to_hstring(manifest.value().displayName));
             }
 
+            auto keyboardAcceleratorKey = VirtualKey::Number1;
```
AnastasiaOrishchenko, line 91: It seems like Windows uses `Ctrl+Shift+0` for some input language settings, so this keyboard event doesn't reach the application. It can be bypassed by changing language settings, but we should probably not force users to do that.    
Combinations of `Ctrl`, `Shift` and `1-9` work fine.

tido64: We should change macOS as well to make it consistent.

windows/ReactTestApp/MainPage.cpp
```diff
@@ -82,13 +88,37 @@ namespace winrt::ReactTestApp::implementation
...
+                    */
+                    keyboardAccelerator.Modifiers(static_cast<VirtualKeyModifiers>(
+                        static_cast<uint32_t>(VirtualKeyModifiers::Control) |
+                        static_cast<uint32_t>(VirtualKeyModifiers::Shift)));
+
+                    keyboardAccelerator.Key(keyboardAcceleratorKey);
+
+                    newMenuItem.KeyboardAccelerators().Append(keyboardAccelerator);
+                    keyboardAcceleratorKey =
+                        static_cast<VirtualKey>(static_cast<uint32_t>(keyboardAcceleratorKey) + 1);
```
tido64, line 119: `VirtualKey` is defined as `enum class VirtualKey : int32_t`, so technically, we should cast to `int32_t`.
```suggestion
                        static_cast<VirtualKey>(static_cast<int32_t>(keyboardAcceleratorKey) + 1);
```
The safer alternative is to use [`std::underlying_type_t`](https://en.cppreference.com/w/cpp/types/underlying_type), but I think this is fine as is.

example/app.json
```diff
@@ -5,6 +5,10 @@
     {
       "appKey": "Example",
       "displayName": "App"
+    },
+    {
+      "appKey": "Example",
+      "displayName": "Also app"
```
tido64, line 11: Revert?

windows/ReactTestApp/MainPage.cpp
tido64: https://github.com/microsoft/react-native-test-app/pull/195

PR #199, opened 2020-08-27
ci: Add Windows target

AnastasiaOrishchenko:
- Add manifest tests
- Add Windows CI target which generates, builds and tests Example app


.github/workflows/build.yml
```diff
@@ -281,6 +281,51 @@ jobs:
             ../scripts/xcodebuild.sh macos/TemplateExample.xcworkspace build
           fi
         working-directory: template-example
+  windows:
+    name: "Windows"
+    runs-on: windows-latest
+    strategy: 
+      matrix:
+        configuration: [Debug, Release]
+        platform: [ARM, x64, x86]
```
tido64, line 290: I think we can skip x86...
```suggestion
        platform: [ARM, x64]
```

.github/workflows/build.yml
```diff
@@ -281,6 +281,51 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install
+        run: |
+          yarn
+        working-directory: example
+      - name: Build bundle and create solution
+        run: |
+          yarn build:windows
+          node ../windows/test-app.js
```
tido64, line 309: We should use the installed script to ensure we didn't make any wrong assumptions:
```suggestion
          node node_modules/react-native-test-app/windows/test-app.js
```

.github/workflows/build.yml
```diff
@@ -281,6 +281,51 @@ jobs:
...
+        working-directory: example
+      - name: Install NuGet packages
+        run: |
+          nuget restore
+        working-directory: example/windows
+      - name: Check sln file
+        run: |
+          type ReactTestApp.sln
+        working-directory: example/windows
```
tido64: Is this still needed?

.github/workflows/build.yml
```diff
@@ -281,6 +281,51 @@ jobs:
...
+      - name: Set up Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install
+        run: |
+          yarn
+        working-directory: example
```
tido64, line 305: Should solution generation and `nuget restore` be part of this step?

.github/workflows/build.yml
```diff
@@ -281,6 +281,47 @@ jobs:
...
+      - name: Build bundle and create solution
+        run: |
+          yarn build:windows
```
tido64, line 308: We should move `yarn build:windows` to the "Build" step.

windows/ReactTestApp/Manifest.cpp
```diff
@@ -91,9 +91,9 @@ namespace ReactTestApp
         m.components = j.at("components").get<std::vector<Component>>();
     }
 
-    std::optional<Manifest> GetManifest()
+    std::optional<Manifest> GetManifest(std::string manifestFileName)
```
tido64, line 94:
```suggestion
    std::optional<Manifest> GetManifest(std::string const &manifestFileName)
```

windows/test-app.js
```diff
@@ -1,7 +1,6 @@
 //@ts-check
 const path = require("path");
 const fs = require("fs");
-const chalk = require("chalk");
```
tido64, line 4: Was it intentional to remove colors?

windows/test-app.js
```diff
@@ -96,6 +95,32 @@ function copyAndReplace(
...
+    "\\$\\(ReactTestAppProjectPath\\)": path.relative(
+      testProjectFilePath,
+      projectFilesDestPath
+    ),
+  };
+
+  copyAndReplace(
+    path.join(testProjectFilePath, testProjectTemplateFile),
+    testProjectFilePath,
+    "ReactTestAppTests.vcxproj",
```
tido64, line 119: Can we parameterize this? Maybe we should split this in two PRs? One for adding the CI, the second to add tests?

windows/ReactTestApp.sln
```diff
@@ -29,6 +29,7 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Mso", "$(ReactNativeModuleP
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Shared", "$(ReactNativeModulePath)\Shared\Shared.vcxitems", "{2049DBE9-8D13-42C9-AE4B-413AE38FFFD0}"
 EndProject
+$(TestProject)
```
tido64, line 32: What happens with this variable if the test project doesn't exist?

tido64 (changes requested):

windows/ReactTestApp.sln
AnastasiaOrishchenko: It's replaced by empty string. 
It could be an option to remove this completely and build test project separately from the solution, but it would remove possibility to easily run tests with Test Explorer in VS.

tido64: This is perfectly fine. I didn't see that it would be replaced with an empty string.

.github/workflows/build.yml
AnastasiaOrishchenko: If we move `yarn build:windows` to "Build" step and solution generation to install, then bundle assets will not be included in the project. It doesn't really matter in terms of CI right now though.

tido64: Ah, you're right.

tido64 (approved):

PR #202, opened 2020-08-28
Manifest tests

AnastasiaOrishchenko:
- Add manifest tests
- Add test step to Windows CI target


example/windows/ReactTestAppTests/ManifestTests.cpp
```diff
@@ -0,0 +1,96 @@
...
+{
+	TEST_CLASS(ManifestTests)
+	{
+	public:
+		TEST_METHOD(ParseManifestWithOneComponent)
+		{
+            std::optional<ReactTestApp::Manifest> manifest = ReactTestApp::GetManifest("manifestTestFiles/simpleManifest.json");
+            if (!manifest.has_value()) {
+                Assert::Fail(L"Couldn't read manifest file");
+            }
```
tido64, line 20: Can you run `clang-format` on this file?

example/windows/ReactTestAppTests/ManifestTests.cpp
```diff
@@ -0,0 +1,96 @@
...
+        {
+            std::optional<ReactTestApp::Manifest> manifest =
+                ReactTestApp::GetManifest("manifestTestFiles/withMultipleComponents.json");
+            if (!manifest.has_value()) {
+                Assert::Fail(L"Couldn't read manifest file");
+            }
+            auto &manifestContents = manifest.value();
+            Assert::IsTrue(manifestContents.name == "Example");
+            Assert::IsTrue(manifestContents.displayName == "Example");
+            Assert::IsTrue(manifestContents.components.size() == 2);
```
tido64, line 39: Should we be using the more descriptive asserts like `AreEqual` etc? https://docs.microsoft.com/en-us/visualstudio/test/microsoft-visualstudio-testtools-cppunittestframework-api-reference?view=vs-2019#cppUnitTestAssert_h

windows/test-app.js
```diff
@@ -207,6 +230,31 @@ function copyProjectTemplateAndReplace(
   }
 
   const solutionFileDestPath = path.join(destPath, windowsDir);
```
tido64, line 232: We should allow user specified output directory (#203)

windows/test-app.js
```diff
@@ -207,6 +230,31 @@ function copyProjectTemplateAndReplace(
...
+
+  const testProjectFilePath = path.join(
+    solutionFileDestPath,
+    "ReactTestAppTests"
+  );
+  const testProjectTemplateFile = "ReactTestAppTestsTemplate.vcxproj";
+  let testProjectEntry = "";
+
+  if (fs.existsSync(path.join(testProjectFilePath, testProjectTemplateFile))) {
```
tido64, line 241: This is a bit weird check. Should we be checking whether `ReactTestAppTests\ReactTestAppTests.vcxproj` exists rather than `ReactTestAppTestsTemplate.vcxproj`?

windows/test-app.js
```diff
@@ -207,6 +230,31 @@ function copyProjectTemplateAndReplace(
...
+    testProjectEntry =
+      'Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "ReactTestAppTests", "ReactTestAppTests\\ReactTestAppTests.vcxproj", "{D2B221C0-0781-4D20-8BF1-D88684662A5D}"\n' +
```
tido64, line 243: Allow user specified test project (#204)

example/windows/ReactTestAppTests/ManifestTests.cpp
AnastasiaOrishchenko: For some reason `AreEqual` is not so happy about comparing `char *` which is type of expected value and `std::string` which is type of actual value without explicit conversion. We could do something like this if it's still readable
`Assert::AreEqual(manifestContents.name, std::string("Example"));`

tido64: Can we try using [`std::string_view`](https://en.cppreference.com/w/cpp/string/basic_string_view)? This requires the tests to be compiled with C++17 enabled:
```c++
#include <string_view>
...
using namespace std::literals::string_view_literals;
...
Assert::AreEqual(manifestContents.name, "Example"sv);
```

windows/test-app.js
AnastasiaOrishchenko: Yes, I agree that it's weird. The reason we do it is that there are a couple of paths that need to be dynamically resolved in `ReactTestAppTests.vcxproj`, so what happens here is that it gets created from the template. If you have any suggestions on how to improve it, I would appreciate it.

example/windows/ReactTestAppTests/ManifestTests.cpp
tido64: Alternatively, I think the compiler should also be able to infer that you're calling `std::string` constructor with just `{"Example"}`.

windows/test-app.js
tido64: Let's leave this as is for now. I want to improve the overall story for adding test projects in #204.

example/windows/ReactTestAppTests/ManifestTests.cpp
AnastasiaOrishchenko: `std::string_view` doesn't work for me with C++17 enabled. `{"Example"}` works fine.

windows/ReactTestApp/pch.h
```diff
@@ -17,5 +17,8 @@
 #include <winrt/Windows.UI.Xaml.Input.h>
 #include <winrt/Windows.UI.Xaml.Interop.h>
 #include <winrt/Windows.UI.Xaml.Markup.h>
+#include <winrt/Windows.UI.Xaml.Input.h>
 #include <winrt/Windows.UI.Xaml.Navigation.h>
 #include <winrt/Windows.UI.Xaml.h>
+#include <winrt/Windows.UI.Input.Preview.Injection.h>
+
```
tido64, line 24: Revert this file? Looks like a bad merge.

tido64 (changes requested):

example/windows/.gitignore
```diff
@@ -0,0 +1,24 @@
...
+*.user
+*.sln.docstates
+
+# Build results
+[Dd]ebug/
+[Dd]ebugPublic/
+[Rr]elease/
+[Rr]eleases/
+AppPackages/
+x64/
```
tido64, line 14: Should we also ignore `ARM`?
```suggestion
AppPackages/
ARM/
x64/
```

example/windows/ReactTestAppTests/ManifestTests.cpp
```diff
@@ -0,0 +1,101 @@
+#include "pch.h"
+
+#include <string>
+
+#include "CppUnitTest.h"
```
tido64, line 5: Is this a third-party header? We should use angle brackets then.
```suggestion
#include <CppUnitTest.h>
```

tido64 (approved):
Only a few nitpicks, otherwise LGTM 👍

tido64 (approved):

PR #205, opened 2020-08-31
Plop template windows

### LINKED ISSUE #203: Windows: Allow user specified output directory for generated solution
`test-app.js` currently always outputs to `./windows/ReactTestApp.sln`. We should allow it to be customizable, e.g. by specifying `--project-directory=windows/`.

AnastasiaOrishchenko:
- Add windows target to plop template
- Add windows-template target to CI (should be slightly changed when windows folder is part of released npm package)
- Resolves #203 to some extent 

From `example` folder, we can now run `yarn install-windows-test-app --projectDirectory=windows to generate the solution.


plopfile.js
```diff
@@ -172,6 +184,9 @@ module.exports = (plop) => {
                 ...(includeMacOS
                   ? { "react-native-macos": "0.62.1" }
                   : undefined),
+                ...(includeWindows
+                  ? { "react-native-windows": "0.62.2" }
```
tido64, line 188:
```suggestion
                  ? { "react-native-windows": "0.62.6" }
```

example/windows/test-app.bat
```diff
@@ -0,0 +1 @@
+node node_modules/react-native-test-app/windows/test-app.js "windows/" "Example"
```
tido64, line 1: Instead of putting this `.bat` file here, we should add the script under `bin` in `react-native-test-app`'s `package.json`. See https://docs.npmjs.com/files/package.json#bin.

Something like:
```json
  "bin": {
    "install-windows-test-app": "windows/test-app.bat"
  }
```

.github/workflows/build.yml
```diff
@@ -323,21 +323,56 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install
+        run: |
+          yarn link #Remove when windows folder is part of release 
+          scripts/install-test-template.sh windows
+        shell: bash
+      - name: Build bundle and create solution
+        run: |
+          yarn link react-native-test-app #Remove when windows folder is part of release
```
tido64, line 364: We should be able to remove `yarn link ...` now, right?

.github/workflows/build.yml
```diff
@@ -323,21 +323,56 @@ jobs:
...
+          yarn build:windows
+          ./test-app
+        working-directory: template-example
+      - name: Install NuGet packages
+        run: |
+          nuget restore
+        working-directory: template-example
+      - name: Build
+        run: |
+          MSBuild TemplateExample.sln /t:Rebuild /p:Configuration=${{ matrix.configuration }} /p:Platform=${{ matrix.platform }}
```
tido64, line 374: Can we try adding [`/m`](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-command-line-reference?view=vs-2019) to see if it speeds up the build?

windows/test-app.js
```diff
@@ -1,18 +1,37 @@
...
+
 //@ts-check
 const path = require("path");
 const fs = require("fs");
 
-const windowsDir = "windows";
+const { argv } = require("yargs")
+  .usage("Usage: $0 --projectDirectory=directory --applicationName=name")
+  .option("projectDirectory", {
+    alias: "projectDirectory",
```
tido64, line 10: Shouldn't the alias be a shorter version of the flag? E.g. `"p"`

tido64: (Same for `"applicationName"`)

windows/test-app.js
```diff
@@ -1,18 +1,37 @@
+#!/usr/bin/env node
+
 //@ts-check
 const path = require("path");
 const fs = require("fs");
 
-const windowsDir = "windows";
+const { argv } = require("yargs")
+  .usage("Usage: $0 --projectDirectory=directory --applicationName=name")
```
tido64, line 8: Can we read `applicationName` from `app.json`?

AnastasiaOrishchenko: The thought behind this was to use app name provided by user when running plop (by running `test-app.js` with correct parameters from `test-app.bat`). But that wouldn't work if we have `bin` script which runs `test-app.js` directly. So sure, we can use `applicationName` from the manifest.

.github/workflows/build.yml
```diff
@@ -323,21 +323,54 @@ jobs:
...
           MSBuild ReactTestAppTests.vcxproj /t:Build /p:Configuration=${{ matrix.configuration }} /p:Platform=${{ matrix.platform }}
           if ("${{ matrix.platform }}" -eq "x64") { VSTest.Console.exe ${{ matrix.platform }}\${{ matrix.configuration }}\ReactTestAppTests.dll }
         working-directory: example/windows/ReactTestAppTests
+  windows-template:
+    name: "Windows [template]"
+    runs-on: windows-latest
+    strategy: 
+      matrix:
+        configuration: [Debug, Release]
+        platform: [ARM, x64]
```
tido64, line 347: I don't think we need to build all configurations again for the template builds. The template builds are more focused on whether we can build windows when 1) the user chose all platforms or 2) the user chose Windows only when asked.
```suggestion
        template: [all, windows]
```

windows/test-app.js
```diff
@@ -53,24 +72,18 @@ const binaryExtensions = [".png", ".pfx"];
...
-  const fullDestPath = path.join(destPath, relativeDestPath);
+function copyAndReplace(srcPath, destPath, replacements = {}) {
   if (fs.lstatSync(srcPath).isDirectory()) {
-    if (!fs.existsSync(fullDestPath)) {
-      fs.mkdirSync(fullDestPath);
+    if (!fs.existsSync(destPath)) {
+      fs.mkdirSync(destPath);
     }
     return;
   }
```
tido64, line 81: Do we still need this? I don't see that we call `copyAndReplace()` with a directory.

windows/test-app.js
AnastasiaOrishchenko: Would this mean that if `app.json` doesn't exist, we don't create solution at all?

tido64: Hmm… Can we output a warning and fall back to "ReactTestApp.sln"?

plopfile.js
```diff
@@ -338,6 +353,45 @@ module.exports = (plop) => {
...
+      if (!exclusive || platforms === "windows") {
+        if (isInstalled("react-native-windows", exclusive)) {
+          const prefix = exclusive ? "" : "windows/";
+          actions.push({
+            type: "add",
+            path: `${prefix}test-app.bat`,
+            template: `node ./node_modules/react-native-test-app/windows/test-app.js --projectDirectory=${
+              exclusive ? "./" : "windows/"
+            }`,
+          });
```
tido64, line 365: Do we still need this?

tido64: Should maybe add this to `postinstall` in `package.json`?

tido64: On the other hand, `resources` will always be empty, so maybe not. Regardless, I don't think we need the `.bat` file is the script is installed.

windows/test-app.js
```diff
@@ -1,18 +1,30 @@
...
 const fs = require("fs");
 
-const windowsDir = "windows";
+const { argv } = require("yargs")
+  .usage("Usage: $0 --projectDirectory=directory")
+  .option("projectDirectory", {
+    alias: "p",
+    type: "string",
+    description: "Directory where solution will be created",
+    default: "./",
```
tido64, line 13: Should we default to `./windows`? I think that's a more common scenario.

windows/test-app.js
```diff
@@ -149,53 +147,61 @@ function copyProjectTemplateAndReplace(
...
           );
+          if (fs.statSync(resourceSrcPath).isDirectory()) {
+            relativeResourcePath = relativeResourcePath.concat("\\**\\*");
+            bundleDirContent = bundleDirContent.concat(
+              relativeResourcePath + ";"
+            );
+          } else {
+            bundleFileContent = bundleFileContent.concat(
+              relativeResourcePath + ";"
+            );
```
tido64, line 190: Nit: `concat()` allows specifying more strings, so we should probably take advantage of that:
```suggestion
            bundleDirContent = bundleDirContent.concat(
              relativeResourcePath, "\\**\\*;"
            );
          } else {
            bundleFileContent = bundleFileContent.concat(
              relativeResourcePath, ";"
            );
```

tido64 (approved):
LGTM; just a few nitpicks :+1:

tido64 (approved):

PR #213, opened 2020-09-08
fix(Android): Unable to disable Flipper

tido64:


android/test-app-util.gradle
```diff
@@ -83,7 +83,8 @@ ext.getFlipperVersion = { baseDir ->
 
     // Prefer user specified Flipper version
     if (project.hasProperty('FLIPPER_VERSION')) {
-        return project.getProperty('FLIPPER_VERSION')
+        def flipperVersion = project.getProperty('FLIPPER_VERSION')
+        return flipperVersion == "false" ? false : flipperVersion
```
arazabishov, line 87: It makes me uncomfortable to return booleans and integers from the same function. Can we return null instead?

arazabishov (changes requested):

arazabishov (approved):

PR #214, opened 2020-09-08
fix(Android): RedBox is dismissed on app start

tido64:
If there is only one component, we navigate to it regardless of whether React Native was successfully initialized. This dismisses any RedBox instances, making it hard to debug errors.


android/app/src/main/java/com/react/testapp/MainActivity.kt
```diff
@@ -51,9 +52,12 @@ class MainActivity : ReactActivity() {
     }
 
     private val startComponentActivity = { component: ComponentViewModel ->
+        didInitialNavigation = true;
         startActivity(newComponentActivityIntent(component))
     }
 
+    private var didInitialNavigation = false;
```
arazabishov, line 59: This flag is going to be reset when activity is re-created (i.e. configuration changes). We need to store/restore it in the bundle if we want to preserve the value.

arazabishov (changes requested):

arazabishov (approved):

PR #228, opened 2020-09-29
fix: Align with Expo's way to set applicationId/bundleIdentifier

tido64:
From Expo's documentation:
- [android.package](https://docs.expo.io/versions/v39.0.0/config/app/#package)
- [ios.bundleIdentifier](https://docs.expo.io/versions/v39.0.0/config/app/#bundleidentifier)

#### Test Plan

- **Android:**
  1. Setting `TEST_APP_IDENTIFIER` in `gradle.properties` should output a deprecation warning, but still set `applicationId` correctly
  2. Setting `android.package` in `app.json` should correctly set `applicationId`
  3. If both are set, the one in `app.json` should take precedence
- **iOS:**
  1. Calling `test_app_bundle_identifier()` should output output a deprecation warning, but still set `bundleIdentifier` correctly
  2. Setting `ios.bundleIdentifier` in `app.json` should correctly set `bundleIdentifier`
  3. If both are set, the one in `app.json` should take precedence
- **macOS:** See iOS steps.

### DIFF
--- a/android/react-native-build.gradle
+++ b/android/react-native-build.gradle
@@ -4,7 +4,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.6.2'
+        classpath 'com.android.tools.build:gradle:3.6.4'
         classpath 'de.undercouch:gradle-download-task:4.0.4'
     }
 }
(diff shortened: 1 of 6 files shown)


alloy (approved):
The iOS/macOS and Ruby parts LGTM 👍

sweggersen (approved):

PR #229, opened 2020-09-30
Rename Android test app package to com.microsoft.reacttestapp

sweggersen:
Rename Android test app package from `com.react.testapp` to `com.microsoft.reacttestapp` to align with iOS.


arazabishov (approved):

android/app/src/androidTest/java/com/react/testapp/ManifestTest.kt
```diff
@@ -1,8 +1,8 @@
-package com.react.testapp
+package com.microsoft.reacttestapp
```
tido64, line 1: Do we need to rename the folders as well to match?

tido64 (approved):
Please make sure to prefix the commit message with `fix(Android): ` so it gets published.

android/app/src/androidTest/java/com/react/testapp/ManifestTest.kt
tido64: We need to rename `androidTest/java/com/react/testapp/` and `flipper/java/com/react/testapp/` as well

sweggersen: Fixed

PR #230, opened 2020-10-02
React Native Path from Manifest

prashant1207:
**Objective**

To support scenarios where there is custom react-native fork, we need to have configurable way to get react-native path.

**Changes**
We have added a new property in platform specific configuration in `manifest.json`, similar to `bundleIdentifier`.
```
{
  "name": "Example",
  "displayName": "Example",
  "components": [
    {
      "appKey": "Example",
      "displayName": "App"
    }
  ],
  "macos": {
    "bundleIdentifier": "..."
    "reactNativePath": "react-native-macos"
  }
}
```

If the **app manifest** has the property `reactNativePath`, we will use that as react-native path, else we fallback to default value.

### DIFF
--- a/ios/test_app.rb
+++ b/ios/test_app.rb
@@ -48,7 +48,7 @@ def react_native_path_from_manifest(project_root, target_platform)
   return if config.nil?
 
   react_native_path = config['reactNativePath']
-  return Pathname.new(resolve_module(react_native_path)) if react_native_path.is_a? String
+  return react_native_path if react_native_path.is_a? String
 end
 
 def find_file(file_name, current_dir)
@@ -95,7 +95,7 @@ end
 
 def react_native_path(project_root, target_platform)
   react_native_from_manifest = react_native_path_from_manifest(project_root, target_platform)
-  return react_native_from_manifest unless react_native_from_manifest.nil?
+  return Pathname.new(resolve_module(react_native_from_manifest)) unless react_native_from_manifest.nil?
 
   react_native = case target_platform
                  when :ios then 'react-native'
--- a/ios/test_app.rb
+++ b/ios/test_app.rb
@@ -43,7 +43,7 @@ def bundle_identifier(project_root, target_platform)
   @test_app_bundle_identifier
 end
 
-def react_native_path_from_manifest(project_root, target_platform)
+def react_native_from_manifest(project_root, target_platform)
   config = platform_config(project_root, target_platform)
   return if config.nil?
 
@@ -94,7 +94,7 @@ def project_path(file, target_platform)
 end
 
 def react_native_path(project_root, target_platform)
-  react_native_from_manifest = react_native_path_from_manifest(project_root, target_platform)
+  react_native_from_manifest = react_native_from_manifest(project_root, target_platform)
   return Pathname.new(resolve_module(react_native_from_manifest)) unless react_native_from_manifest.nil?
 
   react_native = case target_platform
--- a/ios/test_app.rb
+++ b/ios/test_app.rb
@@ -95,7 +95,7 @@ end
 
 def react_native_path(project_root, target_platform)
   react_native_from_manifest = react_native_from_manifest(project_root, target_platform)
-  return Pathname.new(resolve_module(react_native_from_manifest)) unless react_native_from_manifest.nil?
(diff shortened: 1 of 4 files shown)


ios/test_app.rb
```diff
@@ -80,13 +92,18 @@ def project_path(file, target_platform)
...
-  react_native = case target_platform
-                 when :ios then 'react-native'
-                 when :macos then 'react-native-macos'
-                 else raise "Unsupported target platform: #{target_platform}"
-                 end
-  Pathname.new(resolve_module(react_native))
+def react_native_path(project_root, target_platform)
+  react_native_from_manifest = react_native_path_from_manifest(project_root, target_platform)
+  if !react_native_from_manifest.nil?
+    react_native_from_manifest
```
tido64: Prefer to return early:
```suggestion
  return react_native_from_manifest if react_native_from_manifest.is_a? String
```

ios/test_app.rb
```diff
@@ -38,6 +38,18 @@ def bundle_identifier(project_root, target_platform)
...
 end
 
+def react_native_path_from_manifest(project_root, target_platform) 
+  manifest_path = find_file('app.json', project_root)
+  unless manifest_path.nil?
+      manifest = JSON.parse(File.read(manifest_path))
+      platform_config = manifest[target_platform]
+      if !platform_config.nil? && !platform_config.empty?
+        reactNativePath = platform_config['reactNativePath']
+        return reactNativePath if reactNativePath.is_a? String
```
tido64, line 48:
```suggestion
        return Pathname.new(reactNativePath) if reactNativePath.is_a? String
```

tido64, line 48: We need to resolve `reactNativePath` to a path relative to where `Podfile` lives.

ios/test_app.rb
```diff
@@ -38,6 +38,18 @@ def bundle_identifier(project_root, target_platform)
   @test_app_bundle_identifier
 end
 
+def react_native_path_from_manifest(project_root, target_platform)
```
tido64, line 41: Can we add tests for this?

tido64 (changes requested):

ios/test_app.rb
```diff
@@ -38,6 +38,18 @@ def bundle_identifier(project_root, target_platform)
   @test_app_bundle_identifier
 end
 
+def react_native_path_from_manifest(project_root, target_platform) 
+  manifest_path = find_file('app.json', project_root)
+  unless manifest_path.nil?
+      manifest = JSON.parse(File.read(manifest_path))
+      platform_config = manifest[target_platform]
+      if !platform_config.nil? && !platform_config.empty?
```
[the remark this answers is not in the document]
prashant1207: ack.

ios/test_app.rb
prashant1207: I added the path resolution using `Pathname.new(resolve_module(reactNativePath)) if reactNativePath.is_a? String`, would it not suffice?

ios/test_app.rb
prashant1207: For adding positive cases for `react_native_path_from_manifest` will require changes in manifest, Do we use any mock manifest while running the test?

ios/test_app.rb
tido64: Can you post what `Podfile.lock` looks like after running `pod install`? We need to make sure we don't get absolute paths in there.

ios/test_app.rb
tido64: Yes, look under `/test/__fixtures__`

ios/test_app.rb
prashant1207: Pushed the podfile.lock.

ios/test_app.rb
prashant1207: Alright, got it. I will add the tests there.

prashant1207: Added test cases.

alloy (approved):
LGTM 👍

tido64 (approved):
Thanks, Prashant 👍

tido64:
Wiki updated: https://github.com/microsoft/react-native-test-app/wiki/Manifest-(app.json)/_compare/c469860ab6beecc40e3f608dcd9fb13d5374483b...05b5c55227bd29aa41a44cfc078f38d091616fe2

PR #232, opened 2020-10-03
fix: React Native CLI plugin

tido64:
Adds new `init-test-app` command to React Native CLI
```
% yarn react-native init-test-app --help
react-native init-test-app

Initializes a new test app project

Options:
  --destination [string]  Path to the directory where the test app should be created (default: ".")
  --name [string]         Display name of the test app (default: "example-test-app")
  --platform [string]     Specific platform to support: all, android, ios, macos, windows (default: "all")
  -h, --help              output usage information
```

### DIFF
--- a/react-native.config.js
+++ b/react-native.config.js
@@ -1,0 +1,90 @@
+//
+// Copyright (c) Microsoft Corporation
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+//
+// @ts-check
+
+/**
+ * @template ConfigT
+ * @typedef {{
+ *   name: string,
+ *   description?: string,
+ *   func: (argv: Array<string>, config: ConfigT, args: {}) => ?Promise<void>,
+ *   options?: Array<{
+ *     name: string,
+ *     description?: string,
+ *     parse?: (val: string) => any,
+ *     default?:
+ *       | string
+ *       | boolean
+ *       | number
+ *       | ((config: ConfigT) => string | boolean | number),
+ *   }>,
+ *   examples?: Array<{
+ *     desc: string,
+ *     cmd: string,
+ *   }>,
+ * }} Command;
+ */
+
+/**
+ * Infers project name by finding and parsing the nearest `package.json`.
+ * @returns {string | undefined}
+ */
+function inferProjectName() {
+  const { findNearest } = require("./windows/test-app");
+  const packageJson = findNearest("package.json");
+  if (packageJson) {
+    const fs = require("fs");
+    try {
+      const fileContent = fs.readFileSync(packageJson, { encoding: "utf8" });
+      const name = JSON.parse(fileContent)["name"];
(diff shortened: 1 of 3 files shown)


react-native.config.js
```diff
@@ -0,0 +1,68 @@
...
+        plop(path.join(__dirname, "plopfile.js"), {
+          destBasePath: args["destination"],
+        })
+          .getGenerator("app")
+          .runActions(args);
+      },
+      options: [
+        {
+          name: "--destination [string]",
+          description: "Path to the destination directory",
```
alloy, line 51: This is more a long-form version of the option name, so perhaps it can be rephrased to something like:

```suggestion
          description: "Path to the directory where the test app should be created",
```

react-native.config.js
```diff
@@ -0,0 +1,68 @@
...
+          default: process.cwd(),
+        },
+        {
+          name: "--name [string]",
+          description: "Name of the test app",
```
alloy, line 56: Where/how does this surface and should the user really consider it?

react-native.config.js
```diff
@@ -0,0 +1,68 @@
...
+          default: "ReactTestApp",
+        },
+        {
+          name: "--platforms [string]",
+          description:
+            "Platforms to add support for: all, android, ios, macos, windows",
```
alloy, line 62: Is this a comma-separated list? If so, it would be good to state that. E.g.

```suggestion
            "Comma-separated list of platforms to support: all, android, ios, macos, windows",
```

alloy (changes requested):

react-native.config.js
alloy: Alternatively, if it is of little importance, can it default to e.g. the name found in the root `package.json`?

react-native.config.js
tido64: This is supposed to be "pick one" option. Any suggestions on how to phrase that?

alloy: I see. I’d rename the option to be singular and perhaps say something like:

```suggestion
            "Specific platform to support: all, android, ios, macos, windows",
```

Having said that, if you _can_ target `all`, then shouldn’t you be able to pick and permutation too?

(Also, is there a way to infer the permutation from the configuration of the repo?)

react-native.config.js
tido64: You mean like `my-component` → `my-component-test-app`? I think that's a great idea. It's not very important and is mostly used as display name.

tido64: Updated. Let me know what you think.

react-native.config.js
tido64: > Having said that, if you _can_ target `all`, then shouldn’t you be able to pick any permutation too?
> (Also, is there a way to infer the permutation from the configuration of the repo?)

Created issue #233.

I think it's a good idea. We could probably use `react-native config` to detect supported platforms. But I'm hesitant to make such a change in this PR. For now, I think "all" is a reasonable default.

alloy (approved):
Great 👌

:shipit:

PR #237, opened 2020-10-22
fix: Suppress nullability-completeness for React

asmundg:
Newer React Native versions have incomplete nullability annotations,
causing build errors when -Wall is enabled. This change disables these
annotation warnings for the React pods.


tido64 (approved):
Thanks ❤️

PR #238, opened 2020-10-23
fix(macOS): Show where the React apps can be found

tido64:
Adds a screenshot of the React menu as app window background.

image


Saadnajmi:
Bonus points if you have a different image for light and dark mode :P

tom-un (approved):

PR #240, opened 2020-11-03
fix(Apple): Also create an xcscheme to match app name

### LINKED ISSUE #239: Apple: Name the scheme something other than "ReactTestApp"
This came up as a minor detail in one of my PR's (https://github.com/microsoft/fluentui-react-native/pull/529/files#r509388900) that we must use the scheme name and have no control over it. In my scenario, we were adding steps in our build pipeline to build our test app for the purpose of packaging some native dependencies. In our yml, we had to use "ReactTestApp" instead of something more specific to our project, like "FURNTestApp". 

Minor detail, and definitely does not block anything =)

tido64:
Resolves #239

### DIFF
--- a/ios/test_app.rb
+++ b/ios/test_app.rb
@@ -15,6 +15,20 @@ def assert(condition, message)
   raise message unless condition
 end
 
+def app_manifest(project_root)
+  manifest_path = find_file('app.json', project_root)
+  return if manifest_path.nil?
+
+  JSON.parse(File.read(manifest_path))
+end
+
+def app_name(project_root)
+  manifest = app_manifest(project_root)
+  return [nil, nil] if manifest.nil?
+
+  [manifest['name'], manifest['displayName']]
+end
+
 def autolink_script_path
   package_path = resolve_module('@react-native-community/cli-platform-ios')
   File.join(package_path, 'native_modules')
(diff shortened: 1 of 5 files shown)


ios/test_app.rb
```diff
@@ -201,10 +214,15 @@ def make_project!(xcodeproj, project_root, target_platform)
...
-  FileUtils.cp(File.join(src_xcodeproj, 'project.pbxproj'), dst_xcodeproj)
-  FileUtils.ln_sf(File.join(src_xcodeproj, 'xcshareddata'), dst_xcodeproj)
+  # Copy Xcode project files
+  FileUtils.mkdir_p(destination)
+  FileUtils.cp_r(src_xcodeproj, destination)
+  name, _display_name = app_name(project_root)
+  unless name.nil?
+    xcschemes_path = File.join(dst_xcodeproj, 'xcshareddata', 'xcschemes')
+    FileUtils.cp(File.join(xcschemes_path, 'ReactTestApp.xcscheme'),
+                 File.join(xcschemes_path, "#{name}.xcscheme"))
```
tido64, line 224: @Saadnajmi: Would this be enough? It just creates a copy of the scheme. Doesn't change the installed app name or anything. We could address that in a separate PR if desired.

Saadnajmi: @anandrajeswaran since he's the one who suggested it

Does that mean that there is 2 identical schemes now? ReactTestApp and <app name> ?

anandrajeswaran: SHouldn't it just create one scheme that is the {name} of the app?

tido64: I'm contemplating whether to keep `ReactTestApp` to be backwards compatible. It doesn't hurt, I think. And we can remove it in a later version.

Saadnajmi: That's fine by me to not have breaking changes. Thanks for actually implementing the feature request!

alloy (approved):
LGTM 👍

PR #245, opened 2020-11-13
chore(Apple): Use SwiftFormat

tido64:
### Description

Adds a new command, `format:swift`, to format Swift files using SwiftFormat so we no longer have to discuss how to format Swift files. It is also added to the CI so we can get annotations in code review.

See also ADO-364860.

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [ ] Windows

### Test plan

- I deliberately put in something that SwiftFormat should trigger on. [Update: Screenshot posted below.]
- The code is formatted but shouldn't change anything functionally. All CI should still pass.

### DIFF
--- a/.github/swiftformat.json
+++ b/.github/swiftformat.json
@@ -1,0 +1,18 @@
+{
+  "problemMatcher": [
+    {
+      "owner": "swiftformat",
+      "pattern": [
+        {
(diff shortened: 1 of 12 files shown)


tido64:
Build annotation:
image

arazabishov (approved):

PR #250, opened 2020-11-23
fix(Windows): Test app fails to load embedded JS bundle

tido64:
### Description

The test app on Windows doesn't render anything when loading the embedded JS bundle due to incompatible settings. This change ensures that those settings can not be enabled while the embedded bundle is loaded.

ADO-367196

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [x] Windows

### Test plan

1. Build the test app for Windows
    ```sh
    cd example
    yarn
    yarn build:windows
    yarn install-windows-test-app
    start windows/Example.sln
    ```
2. Run the test app without the dev server
3. Try switching between embedded and dev server
4. Run the test app with the dev server (`yarn start:windows`)
5. Try switching between embedded and dev server

| Before | After |
|:-:|:-:|
| image | image |

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,5 +7,6 @@
         "optional": "cpp",
         "string": "cpp",
         "vector": "cpp"
-    }
+    },
+    "react-native-tools.projectRoot": "./example"
 }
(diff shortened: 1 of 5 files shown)


arazabishov (approved):

sweggersen (approved):
LGTM

PR #260, opened 2020-12-15
fix(iOS): Images don't render on iOS 14 with RN < 0.63.2

tido64:
### Description

This change monkey patches React Native on versions prior to 0.63.2 to
fix a bug with images not rendering when on iOS 14.

See https://github.com/facebook/react-native/pull/29420.

### Platforms affected

- [ ] Android
- [x] iOS
- [ ] macOS
- [ ] Windows

### Test plan

1. Downgrade Example app to 0.62
    ```diff
    diff --git a/example/package.json b/example/package.json
    index 351eac3..3a1bff6 100644
    --- a/example/package.json
    +++ b/example/package.json
    @@ -24,10 +24,10 @@
       "devDependencies": {
         "@babel/core": "^7.0.0",
         "mkdirp": "^0.5.1",
    -    "react": "16.13.1",
    -    "react-native": "0.63.3",
    -    "react-native-macos": "0.63.1",
    +    "react": "16.11.0",
    +    "react-native": "0.62.2",
    +    "react-native-macos": "0.62.16",
         "react-native-test-app": "../",
    -    "react-native-windows": "0.63.10"
    +    "react-native-windows": "0.62.17"
       }
     }
    ```
2. Delete `/example/ios/Podfile.lock` before running `pod install`
3. Build Example iOS app

| Before | After |
|:-:|:-:|
| Simulator Screen Shot - iPhone 12 Pro - 2020-12-16 at 09 09 55 | Simulator Screen Shot - iPhone 12 Pro - 2020-12-16 at 09 10 13 |

### DIFF
--- a/ios/ReactTestApp.xcodeproj/project.pbxproj
+++ b/ios/ReactTestApp.xcodeproj/project.pbxproj
@@ -7,15 +7,15 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		1914199A234B2DD800D856AE /* RCTDevSupport+UIScene.m in Sources */ = {isa = PBXBuildFile; fileRef = 19141999234B2DD800D856AE /* RCTDevSupport+UIScene.m */; };
-		192DD201240FCAF5004E9CEB /* Manifest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 192DD200240FCAF5004E9CEB /* Manifest.swift */; };
-		196C22622490CB7600449D3C /* React+Compatibility.m in Sources */ = {isa = PBXBuildFile; fileRef = 196C22602490CB7600449D3C /* React+Compatibility.m */; };
-		196C7215232F1788006556ED /* ReactInstance.swift in Sources */ = {isa = PBXBuildFile; fileRef = 196C7214232F1788006556ED /* ReactInstance.swift */; };
-		196C724123319A85006556ED /* QRCodeReaderDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 196C724023319A85006556ED /* QRCodeReaderDelegate.swift */; };
-		1988284524105BEC005057FF /* UIViewController+ReactTestApp.m in Sources */ = {isa = PBXBuildFile; fileRef = 1988284424105BEC005057FF /* UIViewController+ReactTestApp.m */; };
 		19ECD0D6232ED425003D8557 /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 19ECD0D5232ED425003D8557 /* AppDelegate.swift */; };
-		19ECD0D8232ED425003D8557 /* SceneDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 19ECD0D7232ED425003D8557 /* SceneDelegate.swift */; };
 		19ECD0DA232ED425003D8557 /* ContentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 19ECD0D9232ED425003D8557 /* ContentView.swift */; };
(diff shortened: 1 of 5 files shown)


ios/ReactTestApp/React+Compatibility.m
```diff
@@ -25,3 +48,37 @@ void RTATriggerReloadCommand(RCTBridge *bridge, NSString *reason)
...
+
+#if !TARGET_OS_OSX && REACT_NATIVE_VERSION < 6302
+
+#import <React/RCTUIImageViewAnimated.h>
+
+@implementation RCTUIImageViewAnimated (ReactTestApp)
+
++ (void)initialize
+{
+    if ([self class] != [RCTUIImageViewAnimated class]) {
```
tom-un, line 63: I think initialize can be called more than once if there are other subclasses or categories on RCTUIImageViewAnimated.   Put inside dispatch_once ?

ios/ReactTestApp/React+Compatibility.m
```diff
@@ -25,3 +48,37 @@ void RTATriggerReloadCommand(RCTBridge *bridge, NSString *reason)
...
+    if (@available(iOS 14.0, *)) {
+        RTASwizzleSelector([self class], @selector(displayLayer:), @selector(rta_displayLayer:));
+    }
+}
+
+- (void)rta_displayLayer:(CALayer *)layer
+{
+    if ([self respondsToSelector:@selector(currentFrame)] &&
+        [self performSelector:@selector(currentFrame)] == nil) {
+        [super displayLayer:layer];
```
tom-un, line 76: Took me a minute to figure this out :).   Maybe we should have a comment containing the source of the original displayLayer: method.

The 62 version is
```objc
  if (_currentFrame) {
    layer.contentsScale = self.animatedImageScale;
    layer.contents = (__bridge id)_currentFrame.CGImage;
  }
```
and the post 62 version adds
```objc
else {
    [super displayLayer:layer];
  }
```
so the logic here is doing the opposite of the original `if` and calling the post-62 `else` clause, otherwise call `[self rta_displayLayer:layer]` which I think is intended to call the original RCTUIImageViewAnimated displayLayer.   That's the case for a regular category, but for a swizzle, its not correct because something else *could* have also swizzled the method.   You have to store the `originalMethod` from `RTASwizzleSelector` and call it.   We had a bug in Office where RCT was swizzling a method for the shake gesture, but Office was also swizzling the method, so the RCT swizzle made the Office method no longer callable.   So in the react-native-macos fork I changed the RCTSwizzle method to return the previous selector and store it.   The RTASizzle should do the same so we preserve any chain of swizzles.

tom-un (changes requested):

ios/ReactTestApp/React+Compatibility.m
tido64: Apparently, methods get (unreliably) overridden when redefined in categories: https://stackoverflow.com/questions/5272451/overriding-methods-using-categories-in-objective-c. In which case, the if-check should be enough. I've added `dispatch_once` anyway just to be safe.

ios/ReactTestApp.xcodeproj/project.pbxproj
```diff
@@ -220,8 +220,8 @@
 		19ECD0CA232ED425003D8557 /* Project object */ = {
 			isa = PBXProject;
 			attributes = {
-				LastSwiftUpdateCheck = 1100;
-				LastUpgradeCheck = 1220;
+				LastSwiftUpdateCheck = 1230;
+				LastUpgradeCheck = 1230;
```
kelset, line 224: Are these changes happening because you are using Xcode 12.3? Not relevant per se, just wondering

kelset (commented):
LGTM

tom-un (approved):

ios/ReactTestApp.xcodeproj/project.pbxproj
tido64: Yeah, I had just installed Xcode 12.3 😄

PR #262, opened 2020-12-19
fix: Open last component from previous session

tido64:
### Description

Adds a toggle to reopen the component that was last opened in previous session.

### Platforms affected

- [x] Android
- [x] iOS
- [x] macOS
- [x] Windows

### Test plan

1. Build and run Example app
2. Note that we are seeing a blank window on desktop, and the component list on mobile
3. Enable _Remember last opened component_ (see screenshots below)
4. Open any component
5. Restart Example app
6. The component that was opened last session should now be reopened
7. Disable _Remember last opened component_
8. Restart Example app
9. We should now see either a blank window or the component list
10. Repeat steps 1-9 for all platforms

| Android | iOS | macOS | Windows |
|:-:|:-:|:-:|:-:|
| image | Simulator Screen Shot - iPhone 12 Pro - 2020-12-20 at 11 00 04 | image | image |

### DIFF
--- a/android/app/src/main/java/com/microsoft/reacttestapp/MainActivity.kt
+++ b/android/app/src/main/java/com/microsoft/reacttestapp/MainActivity.kt
@@ -14,7 +14,6 @@ import androidx.recyclerview.widget.DividerItemDecoration
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
 import com.facebook.react.ReactActivity
-import com.facebook.react.bridge.ReactContext
 import com.facebook.react.modules.systeminfo.ReactNativeVersion
 import com.google.android.material.appbar.MaterialToolbar
 import com.microsoft.reacttestapp.component.ComponentActivity
@@ -37,8 +36,7 @@ class MainActivity : ReactActivity() {
     @Inject
     lateinit var bundleNameProvider: ReactBundleNameProvider
 
-    private val testAppReactNativeHost: TestAppReactNativeHost
-        get() = reactNativeHost as TestAppReactNativeHost
+    private var didInitialNavigation = false
 
     private val newComponentViewModel = { component: Component ->
         ComponentViewModel(
@@ -49,8 +47,12 @@ class MainActivity : ReactActivity() {
         )
     }
 
-    private val startComponent: (ComponentViewModel) -> Unit = { component: ComponentViewModel ->
-        didInitialNavigation = true;
+    private val session by lazy {
+        Session(applicationContext)
+    }
+
+    private val startComponent: (ComponentViewModel) -> Unit = { component ->
+        didInitialNavigation = true
         when (component.presentationStyle) {
             "modal" -> {
                 ComponentBottomSheetDialogFragment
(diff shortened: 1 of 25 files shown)


android/app/src/main/java/com/microsoft/reacttestapp/MainActivity.kt
```diff
@@ -72,27 +75,60 @@ class MainActivity : ReactActivity() {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_main)
```
arazabishov, line 77: You need to restore the value of `didInitialNavigation` from the `savedInstanceState` bundle here.

android/app/src/main/java/com/microsoft/reacttestapp/Session.kt
```diff
@@ -0,0 +1,44 @@
+package com.microsoft.reacttestapp
+
+import android.content.Context
+import android.content.Context.MODE_PRIVATE
+import android.content.SharedPreferences
+import androidx.core.content.edit
+
+class Session(context: Context) {
+
+    private companion object {
```
arazabishov, line 10: You don't need to use companion objects for private constants. You can simply extract them from the class to the file level and make them private.

android/app/src/main/java/com/microsoft/reacttestapp/manifest/ManifestProvider.kt
```diff
@@ -10,16 +11,27 @@ class ManifestProvider @Inject constructor(
...
     private val adapter: JsonAdapter<Manifest>
 ) {
-    val manifest: Manifest? by lazy {
+    fun fromResources(): Pair<Manifest?, String> {
         val appIdentifier = context.resources
             .getIdentifier("raw/app", null, context.packageName)
 
-        if (appIdentifier != 0) {
+        return if (appIdentifier == 0) {
+            Pair(null, "")
```
arazabishov, line 19: Instead of passing empty string, make the return type optional instead?

arazabishov (changes requested):

android/app/src/main/java/com/microsoft/reacttestapp/Session.kt
```diff
@@ -0,0 +1,44 @@
...
+        const val REMEMBER_LAST_COMPONENT_ENABLED = "RememberLastComponent/Enabled"
+        const val REMEMBER_LAST_COMPONENT_INDEX = "RememberLastComponent/Index"
+    }
+
+    private val sharedPreferences: SharedPreferences =
+        context.getSharedPreferences(BuildConfig.APPLICATION_ID, MODE_PRIVATE)
+
+    private var lastComponentIndex: Int
+        get() = sharedPreferences.getInt(REMEMBER_LAST_COMPONENT_INDEX, -1)
+        set(index) = sharedPreferences.edit { putInt(REMEMBER_LAST_COMPONENT_INDEX, index) }
```
sweggersen, line 21: After edit, you need to commit changes, ref. https://developer.android.com/reference/android/content/SharedPreferences.Editor.

Same for the other edits below

tido64: We're using the [`SharedPreferences.edit` extension](https://developer.android.com/reference/kotlin/androidx/core/content/package-summary#edit) which should automatically call [`apply`](https://developer.android.com/reference/android/content/SharedPreferences/Editor.html#apply()) when the block exits.

android/app/src/main/java/com/microsoft/reacttestapp/Session.kt
tido64: Getting this:
```
e: /~/android/app/src/main/java/com/microsoft/reacttestapp/Session.kt: (10, 13): Const 'val' are only allowed on top level or in objects
e: /~/android/app/src/main/java/com/microsoft/reacttestapp/Session.kt: (11, 13): Const 'val' are only allowed on top level or in objects
e: /~/android/app/src/main/java/com/microsoft/reacttestapp/Session.kt: (12, 13): Const 'val' are only allowed on top level or in objects
```
Are we fine with not making them consts?

arazabishov: You need to extract them to the file level:

```kotlin
private const val MANIFEST_CHECKSUM = "ManifestChecksum"
private const val REMEMBER_LAST_COMPONENT_ENABLED = "RememberLastComponent/Enabled"
private const val REMEMBER_LAST_COMPONENT_INDEX = "RememberLastComponent/Index"

class Session(context: Context) {
  ...
}
```

tido64: Nice 👍

arazabishov (approved):

PR #264, opened 2020-12-22
ci(Windows): Cache NuGet packages

tido64:
### Description

Cache global NuGet packages folder to improve Windows build times.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [x] Windows

### Test plan

- All CI should pass.
- Windows build times should be improved.


tido64:
Rejecting this as the cache was about 1.7 GB and took well over 2 minutes to download and extract. With the installation time taking over a minute, it doesn't improve much on the current installation time of 4 minutes.

PR #266, opened 2020-12-26
fix(Apple): Allow passing options to use_react_native!

tido64:
### Description

Allow options to be passed directly to `use_react_native!`, such as `:hermes_enabled` introduced in 0.64:

```rb
use_test_app! :hermes_enabled => true
```

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [ ] Windows

### Test plan

1. Navigate to `/example`
2. Delete `ios/Podfile.lock` and `ios/Pods/`
3. Run `yarn`
4. Run `pod install --project-directory=ios`
5. Build `ios/Example.xcworkspace`
6. Repeat steps 1-6 for versions 0.60, 0.61, 0.62, 0.63
    - Use `yarn set-react-version` to set the correct package versions

### DIFF
--- a/ios/use_react_native-0.63.rb
+++ b/ios/use_react_native-0.63.rb
@@ -5,8 +5,13 @@
 # LICENSE file in the root directory of this source tree.
 #
 
-def include_react_native!(react_native:, target_platform:, project_root:, flipper_versions:)
+def include_react_native!(options)
+  react_native, flipper_versions, project_root, target_platform = options.values_at(
+    :path, :rta_flipper_versions, :rta_project_root, :rta_target_platform
+  )
+
   require_relative File.join(project_root, react_native, 'scripts', 'react_native_pods')
+
   use_flipper!(flipper_versions) if target_platform == :ios && flipper_versions
-  use_react_native!(:path => react_native)
+  use_react_native!(options)
 end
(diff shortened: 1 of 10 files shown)


kelset (approved):
LGTM

ios/use_react_native-0.63.rb
```diff
@@ -5,8 +5,13 @@
...
+def include_react_native!(options)
+  react_native, flipper_versions, project_root, target_platform = options.values_at(
+    :path, :rta_flipper_versions, :rta_project_root, :rta_target_platform
+  )
+
   require_relative File.join(project_root, react_native, 'scripts', 'react_native_pods')
+
   use_flipper!(flipper_versions) if target_platform == :ios && flipper_versions
-  use_react_native!(:path => react_native)
+  use_react_native!(options)
```
tom-un, line 16: I'm not a Ruby expert, but: `use_react_native!` now has three args: `project_root`, `target_platform`, `options`.  Doesn't this call need to include all three?

tido64: `use_react_native!` actually only takes in one argument here: https://github.com/facebook/react-native/blob/811ccec74e5b898e9304f86f85f390908f8d317f/scripts/react_native_pods.rb#L6. The arrow (`=>`) syntax we see is short hand for creating the hash map.

tom-un: Oh, I was confusing the local use_react_native! in ios/test_app.rb versus the use_react_native! in react-native scripts/react_native_pods.rb

PR #267, opened 2021-01-05
macOS arm64 support

Saadnajmi:
### Description

Issue 

We don't test that react-native-test-app works for macOS on arm64.

Fix

Add a step in our CI to explicitly build the macOS test app on the arm64 architecture.

### Platforms affected

- [ ] Android
- [ ] iOS
- [x] macOS
- [ ] Windows

### Test plan

Normal CI loop. I also locally checked the executable using `lipo -archs` to make sure arm64 was indeed built.


Saadnajmi:
This builds locally on my Intel Macbook pro on Big Sur in Xcode 12.2, but fails CI. I'm not sure what the difference is between local and CI, but in a few days I'll have access to an M1 macbook air so I'll try building locally on that machine. 
Screen Shot 2021-01-05 at 11 07 42 PM

macos/ReactTestApp/ReactTestApp.debug.xcconfig
```diff
@@ -6,6 +6,7 @@ GCC_DYNAMIC_NO_PIC = NO
 GCC_OPTIMIZATION_LEVEL = 0
 GCC_PREPROCESSOR_DEFINITIONS = DEBUG=1 $(inherited)
 MTL_ENABLE_DEBUG_INFO = INCLUDE_SOURCE
-ONLY_ACTIVE_ARCH = YES
+ONLY_ACTIVE_ARCH = NO
+ARCHS = $(ARCHS_STANDARD) arm64
```
tido64, line 10: Why do we need to set this? If you're building on an M1, `ARCHS_STANDARD` should already include arm64. If it's only for making sure that things build on CI, can we provide them as parameters to `xcodebuild`?

```
xcodebuild ARCHS="$ARCHS_STANDARD arm64" ONLY_ACTIVE_ARCH=NO ...
```

tido64 (changes requested):

macos/ReactTestApp/ReactTestApp.debug.xcconfig
Saadnajmi: We don't need to. I was trying to ensure that arm64 was built by explicitly stating it. I have confirmed ARCH_STANDARD already has arm64 so we can remove this

tido64: If it's only for validation, I think it makes more sense to add the variables in [`/.github/workflows/build.yml`](https://github.com/microsoft/react-native-test-app/blob/6f64e208e02fbd09b00201a64cadb70a24262f2c/.github/workflows/build.yml#L273). Adding them here means that devs always have to build both archs even though they only need one.

Saadnajmi: Yeah sorry. I agree that this is better not as an xcconfig. Having this draft PR be public is mostly to iterate and test CI since that seemed the simplest way to go about it. 

Some context for what I was trying to do:
I was trying to make sure FluentUI React Native component static libraries were built as universal binaries. Those are built as a by product of building their test app, which is coming from RNTA. As such, I wanted to make sure RNTA was built into a universal binary (I.E build both architectures). That is done mostly by adding `ONLY_ACTIVE_ARCH=NO` to the xcconfig, since `ARCHS_STANDARD` already includes arm64 with Xcode 12.2

If we would rather ReactTestApp not by default always build both architectures, I could look into changing that on the FluentUI React Native side (Build both architectures separately and combine them). So far I have indeed confirmed that ReactTestApp does compile for arm64 locally, despite it failing CI right now.

.github/workflows/build.yml
```diff
@@ -270,7 +270,7 @@ jobs:
           set -eo pipefail
           yarn build:macos || yarn build:macos
           pod install --project-directory=macos
-          ../scripts/xcodebuild.sh macos/Example.xcworkspace build-for-testing
```
Saadnajmi, line 273: Not a permanent change, but removing this action allowed this command to pass locally for me. I wanted to confirm that is true if I passed it to CI.

tido64: Would it make sense to add another step to build arm64?

macos/ReactTestApp/ReactTestApp.debug.xcconfig
Saadnajmi: Talked offline. FluentUI React Native will build a universal binary in it's own pipeline, and this PR will be focused on adding a "Build macOS arm64" step to react-native-test-app's CI

.github/workflows/build.yml
```diff
@@ -276,6 +276,13 @@ jobs:
         run: |
           ../scripts/xcodebuild.sh macos/Example.xcworkspace test-without-building
         working-directory: example
+      - name: Build ARM
+        run: |
+          set -eo pipefail
+          yarn build:macos || yarn build:macos
+          pod install --project-directory=macos
```
tido64, line 283: I don't think we need to do these again:
```suggestion
```

tido64 (changes requested):

.github/workflows/build.yml
Saadnajmi: Added, good catch.

.github/workflows/build.yml
Saadnajmi: Ended up going with this suggestion. Added a step to the macOS job to also build arm64.

PR #272, opened 2021-01-20
fix(iOS): Correctly initialize Flipper with defaults

tido64:
### Description

`SKDescriptorMapper` was initialized without defaults, breaking the
Layout and Logs plugins in Flipper.

### Platforms affected

- [ ] Android
- [x] iOS
- [ ] macOS
- [ ] Windows

### Test plan

CI should pass.


kelset (approved):
this seems to fix both the layout and the logs 👍
Screenshot 2021-01-20 at 15 44 36

Screenshot 2021-01-20 at 15 44 40

PR #273, opened 2021-01-21
ci: Run commitlint to verify commit messages

tido64:
### Description

Ensure the first commit message follows [conventional commit format](https://www.conventionalcommits.org/). GitHub reuses the first commit message when merging against `master` so we should make sure `semantic-release` can process it appropriately.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

- CI should only fail when the first commit message is wrongly formatted

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,30 @@ on:
       - master
   pull_request:
 jobs:
+  lint-commit:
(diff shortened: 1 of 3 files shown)


tido64:
image

arazabishov (approved):

Issue #145, opened 2020-07-02
Feature request: ability to hook `RCTRootView` and/or `RCTBridge` initialization.


tom-un:
It would be useful for consumers to be able to hook into the creation of the `RCTRootView` and/or `RCTBridge` objects with custom code.   For example, if the consumer wanted to run some custom code to build an `initialProperties` NSDictionary to pass to the `RCTRootView` or `RCTBridge` init.   Or if some customization needed to be done to the `RCTRootView` such as changing properties on the view.

tido64:
Closing this as you can create [custom view controllers](https://github.com/microsoft/react-native-test-app/wiki/Platform-Specifics:-iOS,-macOS,-visionOS#using-view-controllers) (iOS) or [activities/fragements](https://github.com/microsoft/react-native-test-app/wiki/Platform-Specifics:-Android#using-fragments) (Android) and perform more complicated setups that way. Feel free to reopen if you strongly disagree.

PR #275, opened 2021-02-02
A mechanism for listening to lifecycle events in RNTA

arazabishov:
### Description

This PR adds a mechanism for listening lifecycle events such as application start, RN init, etc. It does it by looking-up methods in ReactPackage's provided by native modules using reflection.

The issue #145 is somewhat relevant, but doesn't address the exact problem.

### Platforms affected

- [x] Android
- [x] iOS
- [ ] macOS
- [ ] Windows

### Test plan

To test this change, you will need to link a native module that has a ReactPackage with one (or all) of the following methods:

```java
public class LpcAndroidTestReactPackage implements ReactPackage, ReactTestAppLifecycleEvents {
    
    @Override
    public void onTestAppCreated() {
        System.out.println("onTestAppCreated()");
    }

    @Override
    public void onPreInitReactNativeInstance() {
        System.out.println("onPreInitReactNativeInstance()");
    }
    
    ...
}
```

### DIFF
--- a/android/app/src/main/java/com/microsoft/reacttestapp/TestApp.kt
+++ b/android/app/src/main/java/com/microsoft/reacttestapp/TestApp.kt
@@ -29,8 +29,8 @@ class TestApp : Application(), HasAndroidInjector, ReactApplication {
         eventConsumers.forEach { it.onTestAppCreated() }
 
         val testAppComponent = DaggerTestAppComponent.builder()
-                .binds(this)
-                .build()
+            .binds(this)
+            .build()
         testAppComponent.inject(this)
 
         eventConsumers.forEach { it.onPreInitReactNativeInstance() }
--- a/android/app/src/main/java/com/microsoft/reacttestapp/TestApp.kt
+++ b/android/app/src/main/java/com/microsoft/reacttestapp/TestApp.kt
@@ -5,6 +5,7 @@ import com.facebook.react.PackageList
 import com.facebook.react.ReactApplication
 import com.microsoft.reacttestapp.di.DaggerTestAppComponent
 import com.microsoft.reacttestapp.react.TestAppReactNativeHost
+import com.microsoft.reacttestapp.support.ReactTestAppLifecycleEvents
 import dagger.android.AndroidInjector
 import dagger.android.DispatchingAndroidInjector
 import dagger.android.HasAndroidInjector
--- a/android/app/src/main/java/com/microsoft/reacttestapp/TestApp.kt
+++ b/android/app/src/main/java/com/microsoft/reacttestapp/TestApp.kt
@@ -10,16 +10,6 @@ import dagger.android.DispatchingAndroidInjector
 import dagger.android.HasAndroidInjector
 import javax.inject.Inject
 
-private fun PackageList.invokeLifecycleMethod(name: String) {
-    packages.forEach {
-        try {
-            val methodToFind = it.javaClass.getMethod(name)
-            methodToFind.invoke(it)
-        } catch (e: NoSuchMethodException) {
-        }
-    }
-}
-
 class TestApp : Application(), HasAndroidInjector, ReactApplication {
 
     @Inject
@@ -31,15 +21,18 @@ class TestApp : Application(), HasAndroidInjector, ReactApplication {
     override fun onCreate() {
         super.onCreate()
 
-        val packageList = PackageList(this)
-        packageList.invokeLifecycleMethod("onAppCreated")
+        val eventConsumers = PackageList(this).packages
+            .filter { it is ReactTestAppLifecycleEvents }
+            .map { it as ReactTestAppLifecycleEvents }
+
+        eventConsumers.forEach { it.onTestAppCreated() }
 
         val testAppComponent = DaggerTestAppComponent.builder()
                 .binds(this)
                 .build()
         testAppComponent.inject(this)
 
-        packageList.invokeLifecycleMethod("onPreReactNativeInit")
+        eventConsumers.forEach { it.onPreInitReactNativeInstance() }
         reactNativeHostInternal.init()
     }
 
--- a/android/app/src/main/java/com/microsoft/reacttestapp/TestApp.kt
+++ b/android/app/src/main/java/com/microsoft/reacttestapp/TestApp.kt
@@ -1,6 +1,7 @@
 package com.microsoft.reacttestapp
 
 import android.app.Application
+import com.facebook.react.PackageList
 import com.facebook.react.ReactApplication
 import com.microsoft.reacttestapp.di.DaggerTestAppComponent
 import com.microsoft.reacttestapp.react.TestAppReactNativeHost
@@ -8,6 +9,16 @@ import dagger.android.AndroidInjector
 import dagger.android.DispatchingAndroidInjector
 import dagger.android.HasAndroidInjector
 import javax.inject.Inject
+
+private fun PackageList.invokeLifecycleMethod(name: String) {
+    packages.forEach {
+        try {
+            val methodToFind = it.javaClass.getMethod(name)
+            methodToFind.invoke(it)
+        } catch (e: NoSuchMethodException) {
+        }
+    }
+}
 
 class TestApp : Application(), HasAndroidInjector, ReactApplication {
 
@@ -20,12 +31,15 @@ class TestApp : Application(), HasAndroidInjector, ReactApplication {
     override fun onCreate() {
         super.onCreate()
 
-        val testAppComponent = DaggerTestAppComponent.builder()
-            .binds(this)
-            .build()
+        val packageList = PackageList(this)
+        packageList.invokeLifecycleMethod("onAppCreated")
 
+        val testAppComponent = DaggerTestAppComponent.builder()
+                .binds(this)
+                .build()
         testAppComponent.inject(this)
 
+        packageList.invokeLifecycleMethod("onPreReactNativeInit")
         reactNativeHostInternal.init()
     }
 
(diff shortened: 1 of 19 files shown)


example/android/build.gradle
```diff
@@ -1 +1,10 @@
-// Top-level build file where you can add configuration options common to all sub-projects/modules.
+buildscript {
+    repositories {
+        jcenter()
```
tido64, line 3: We should probably start using `mavenCentral()`: https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

arazabishov: I will resolve this for now, as switching to mavenCentral breaks ADP.

ios/ReactTestApp/Public/ReactTestApp-DevSupport.h
```diff
@@ -10,6 +10,7 @@
 NS_ASSUME_NONNULL_BEGIN
 
 extern NSNotificationName const ReactTestAppDidInitializeNotification;
+extern NSNotificationName const ReactTestAppPreInitReactNativeInstanceNotification;
```
tido64, line 13:
```suggestion
extern NSNotificationName const ReactTestAppWillInitializeReactNativeNotification;
```

ios/ReactTestApp/ReactInstance.swift
```diff
@@ -97,6 +97,13 @@ final class ReactInstance: NSObject, RCTBridgeDelegate {
     }
 
     func initReact(onDidInitialize: @escaping () -> Void) {
+        if bridge == nil {
+            NotificationCenter.default.post(
+                name: .ReactTestAppPreInitReactNativeInstance,
+                object: []
```
tido64, line 103: I think you can pass `nil` here:
```suggestion
                object: nil
```

android/app/src/main/java/com/microsoft/reacttestapp/TestApp.kt
```diff
@@ -20,12 +22,18 @@ class TestApp : Application(), HasAndroidInjector, ReactApplication {
...
-            .build()
+        val eventConsumers = PackageList(this).packages
+            .filter { it is ReactTestAppLifecycleEvents }
+            .map { it as ReactTestAppLifecycleEvents }
+
+        eventConsumers.forEach { it.onTestAppCreated() }
 
+        val testAppComponent = DaggerTestAppComponent.builder()
+                .binds(this)
+                .build()
```
tido64, line 33: Did you mean to indent here?
```suggestion
            .binds(this)
            .build()
```

android/build.gradle
```diff
@@ -0,0 +1,10 @@
+buildscript {
```
tido64, line 1: Do we also need this file? We have one in `android/app/` as well...

example/android/build.gradle
```diff
@@ -1 +1,10 @@
-// Top-level build file where you can add configuration options common to all sub-projects/modules.
+buildscript {
```
tido64, line 1: Why do we also need this file now?

test-app.gradle
```diff
@@ -17,8 +17,12 @@ private static void apply(Settings settings) {
     }
 
     settings.include(":app")
+    settings.include(":support")
```
tido64, line 20: Is there no way to leverage autolinking to pull this in? How will the consumer use this? Can we implement an example in the Example app?

tido64 (commented):
Looks fine in general. I left a few questions because I have no idea how Android works.

test-app.gradle
arazabishov: The `support` module will be automatically linked for consumers, so no additional work will be required. To start receiving events, the consumer's `ReactPackage` has to implement the provided interface.

> Can we implement an example in the Example app?

Of course, but we will need to add native modules to iOS/Android examples to implement the interface. I don't remember why we've decided to keep the example as "JS" only.

android/build.gradle
arazabishov: Without linking the ADP here, the project won't be configured correctly. When `android/app` was the only module in the project, it was sufficient to link the plugin just at the module level. Once I've added the `support` module, I had to lift this declaration to the project level, otherwise, two instances of the plugin get into a conflict. 

I should explore if removing the `buildscript` block from the `android/app` can break existing integrations.
 - [ ] Check if the `buildscript` block can be removed safely from the `android/app`.

example/android/build.gradle
arazabishov: Related to the comment above. ADP plugin has to be applied at the project root level (if you have more than one module, app + support). The project root is where your `settings.gradle` file is declared. If you want to be able to develop `RNTA/android` and `RNTA/example/android` in AS independently, they have to be distinct projects.

android/build.gradle
arazabishov: I checked and removed the dupe entry for applying ADP, but the buildscript block still has to be there for the kotlin gradle plugin.

android/app/src/main/java/com/microsoft/reacttestapp/TestApp.kt
```diff
@@ -20,12 +22,18 @@ class TestApp : Application(), HasAndroidInjector, ReactApplication {
...
+        eventConsumers.forEach { it.onTestAppCreated() }
+
         val testAppComponent = DaggerTestAppComponent.builder()
             .binds(this)
             .build()
-
         testAppComponent.inject(this)
 
+        eventConsumers.forEach { it.onPreInitReactNativeInstance() }
         reactNativeHostInternal.init()
```
tido64, line 37: Is there a post init event as well?

android/support/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+android {
+    compileSdkVersion 29
+    buildToolsVersion "29.0.3"
+
+    defaultConfig {
+        minSdkVersion 21
+        targetSdkVersion 29
+        versionCode 1
+        versionName "1.0"
+    }
```
tido64, line 17: Is there any way to keep these in a central place so we don't have to manually sync with `android/app/build.gradle`?

tido64 (approved):
LGTM, only some minor comments.

sweggersen (approved):

android/app/src/main/java/com/microsoft/reacttestapp/TestApp.kt
arazabishov: I took a look at what iOS does, specifically at `ReactTestAppDidInitializeNotification`. It seems that Android's variant of this event has different semantics, or to be more specific, the event is fired after the application start, not after the RN bridge init. I can introduce a new event on Android that is fired after RN initializes itself, but it will be confusing and inconsistent with iOS.

Ideally, events with similar names should have similar semantics on both platforms. I would repurpose the event on iOS to be sent after application start, and add a new event for post-bridge init. This, however, is going to be a breaking change.

Your call @tido64 and @sweggersen.

android/support/build.gradle
arazabishov: Good point, thanks.

android/app/src/main/java/com/microsoft/reacttestapp/TestApp.kt
tido64: Not that many people are using these events besides us, I think. Please go ahead and rename them.

sweggersen: I like the second suggestion. It's better to do these changes sooner than later, and we need to be consistent between platforms

ios/ReactTestApp/Public/ReactTestApp-DevSupport.h
```diff
@@ -10,6 +10,8 @@
 NS_ASSUME_NONNULL_BEGIN
 
 extern NSNotificationName const ReactTestAppDidInitializeNotification;
+extern NSNotificationName const ReactTestAppWillInitializeReactNativeNotification;
+extern NSNotificationName const ReactTestAppInitializedReactNativeNotification;
```
tido64, line 14:
```suggestion
extern NSNotificationName const ReactTestAppDidInitializeReactNativeNotification;
extern NSNotificationName const ReactTestAppWillInitializeReactNativeNotification;
```

tido64 (changes requested):

ios/ReactTestApp/Public/ReactTestApp-DevSupport.h
arazabishov: image

ios/ReactTestApp/AppDelegate.swift
```diff
@@ -23,6 +23,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                      didFinishLaunchingWithOptions _: [UIApplication.LaunchOptionsKey: Any]?) -> Bool
     {
         self.application = application
+
+        NotificationCenter.default.post(
+            name: .ReactTestAppDidInitialize,
+            object: nil
+        )
+
         return true
```
tido64, line 32: Can we put this in a `defer` block so it gets called after returning?
```suggestion
        defer {
            NotificationCenter.default.post(
                name: .ReactTestAppDidInitialize,
                object: nil
            )
        }

        self.application = application
        return true
```

arazabishov: Done. Out curiosity, why?

tido64 (approved):
🚀

ios/ReactTestApp/AppDelegate.swift
tido64: Technically, the app isn't really initialized until after you've returned `true` here.

PR #277, opened 2021-02-03
ci: run release job on Linux to improve perf

tido64:
### Description

The VMs running Ubuntu seem to be consistently faster than the macOS ones when installing dependencies (~1 min vs. ~3 min).

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

CI should pass.


kelset (approved):
interesting!

PR #301, opened 2021-03-30
fix(windows): restore ability to use NuGet

### LINKED ISSUE #299: Windows: Add support for the experimental NuGet feed
[useExperimentalNuget](https://github.com/microsoft/react-native-windows/blob/0ee306b5f13ed42d2cfb4dc54a8bffab374ef03b/packages/%40react-native-windows/cli/src/generator-windows/index.ts#L168) allows downloading of pre-built binaries and significantly reduces build time on Windows.

tido64:
### Description

Restores the ability to use the experimental NuGet packages.

Resolves #299.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [x] Windows

### Test plan

CI should pass.

To test locally, specify the `--use-nuget` flag when installing the Windows test app:
```
cd example
yarn
start windows/Example.sln
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -405,7 +405,7 @@ jobs:
       - name: Build bundle and create solution
         run: |
           yarn build:windows
-          yarn install-windows-test-app
+          yarn install-windows-test-app --use-nuget
         working-directory: example
       - name: Install NuGet packages
         run: |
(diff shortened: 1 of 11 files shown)


tido64:
Looks like there's a significant reduction in build times (template builds don't use NuGet):
image

JasonVMo (approved):
Very cool to see the nuget support added. I assume the 1000.0 versions in the solutions get replaced with the version appropriate entries?

tido64:
> I assume the 1000.0 versions in the solutions get replaced with the version appropriate entries?

Correct. We used an arbitrary version earlier (63.2). I changed it to 1000.0 to make it easier to replace across all files.

PR #312, opened 2021-05-11
fix(android): remove dagger

arazabishov:
### Description

Removing dependency on dagger from the project.

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

Manual validation.


Saadnajmi:
But.. how will you ~~stab yourself in the foot~~ inject dependencies without dagger? 😲

tido64 (approved):

arazabishov:
@Saadnajmi hehe, that's a good one. I have nothing against Dagger, it is just not bringing enough value for RNTA to justify its maintenance cost.

tido64:
@Saadnajmi: And to be completely honest, this change was triggered by the fact that our linters are failing our builds because the Dagger version is too old 😛

PR #323, opened 2021-05-29
fix(android): bump target Android SDK platform to 30

tido64:
### Description

Bumps target Android SDK platform from 29 to 30.

- Android Gradle plugin -> 4.2.0
- Android SDK Build-Tools -> 30.0.3
- Android SDK Platform -> 30
- Gradle -> 6.9
- Kotlin Gradle plugin -> 1.4.30
- com.squareup.moshi:moshi-kotlin -> 1.12.0
- com.squareup.moshi:moshi-kotlin-codegen -> 1.12.0

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

CI should be green.


tido64:
@ArazAbishov, @sweggersen: I had some issues bumping `androidx.appcompat:appcompat` to 1.3.0. It says that `onActivityResult()` is deprecated but I don't know why we were overriding it in the first place. It looks like we only forward the event to fragments? Is it because they don't receive those events otherwise?

https://github.com/microsoft/react-native-test-app/blob/bc3aeae986e80e992000bdac607a70bfd719136f/android/app/src/main/java/com/microsoft/reacttestapp/component/ComponentActivity.kt#L89-L95

arazabishov:
> Is it because they don't receive those events otherwise?

Yep, they were not at the time. It might have changed in the new support libs. I can check that in Midgard (that's where we rely on this event to trickle down to Fragments).

arazabishov (approved):

PR #327, opened 2021-06-04
fix: latest react-native requires cli version >=4.x

tido64:
### Description

Nightlies have been failing because we're preventing the correct version of `@react-native-community/cli` and `metro` to be hoisted.

### Platforms affected

- [x] Android
- [x] iOS
- [x] macOS
- [x] Windows

### Test plan

```sh
yarn set-react-version master
cd example
yarn
yarn build:ios
```


scripts/set-react-version.js
```diff
@@ -33,28 +33,19 @@ const profiles = {
     "react-native-windows": "0.63.14",
   },
   0.64: {
-    "@react-native-community/cli": "^5.0.1-alpha.0",
-    "@react-native-community/cli-platform-android": "^5.0.1-alpha.0",
-    "@react-native-community/cli-platform-ios": "^5.0.1-alpha.0",
```
kelset, line 38: I would argue that we should have >= 5 here...

scripts/set-react-version.js
```diff
@@ -33,28 +33,19 @@ const profiles = {
...
     react: "17.0.1",
     "react-native": "0.64.0",
     "react-native-macos": undefined,
     "react-native-windows": "0.64.3",
   },
   master: {
-    "@react-native-community/cli": "^5.0.1-alpha.0",
-    "@react-native-community/cli-platform-android": "^5.0.1-alpha.0",
-    "@react-native-community/cli-platform-ios": "^5.0.1-alpha.0",
```
kelset, line 47: ...and >= 6 here...

scripts/set-react-version.js
```diff
@@ -33,28 +33,19 @@ const profiles = {
...
-    react: "17.0.1",
+    react: "17.0.2",
     "react-native": "facebook/react-native#master",
     "react-native-macos": undefined,
     "react-native-windows": undefined,
   },
   nightly: {
-    "@react-native-community/cli": "^5.0.1-alpha.0",
-    "@react-native-community/cli-platform-android": "^5.0.1-alpha.0",
-    "@react-native-community/cli-platform-ios": "^5.0.1-alpha.0",
```
kelset, line 56: ...and >= 6 here.

kelset (approved):
my 2c$

scripts/set-react-version.js
tido64: `react-native` pulls in the correct version if we just omit it. That's why I removed it here, and below.

kelset (approved):

PR #348, opened 2021-06-18
fix(windows): crash when no JS bundle is found

tido64:
### Description

Windows test app crashes when trying to load a JS bundle that doesn't exist.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [x] Windows

### Test plan

```sh
cd example
yarn
start windows/Example.sln
yarn start:windows
```

When the test app is loaded, try switching to embedded JS bundle.

Screenshot 2021-06-18 201327

### DIFF
--- a/windows/ReactTestApp/MainPage.cpp
+++ b/windows/ReactTestApp/MainPage.cpp
@@ -155,18 +155,18 @@ IAsyncAction MainPage::LoadFromDevServer(IInspectable const &, RoutedEventArgs)
 void MainPage::LoadFromJSBundle(IInspectable const &, RoutedEventArgs)
 {
     if (!LoadJSBundleFrom(JSBundleSource::Embedded)) {
-        std::string message{
-            "No JavaScript bundle with one of the following names was found in the app:\n\n"};
+        std::wstring message{
+            L"No JavaScript bundle with one of the following names was found in the app:\n\n"};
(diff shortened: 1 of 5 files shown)


windows/ReactTestApp/MainPage.cpp
```diff
@@ -136,14 +138,36 @@ MainPage::MainPage()
...
+    }
+
     LoadJSBundleFrom(JSBundleSource::DevServer);
 }
 
 void MainPage::LoadFromJSBundle(IInspectable const &, RoutedEventArgs)
 {
-    LoadJSBundleFrom(JSBundleSource::Embedded);
+    if (!LoadJSBundleFrom(JSBundleSource::Embedded)) {
+        std::string message{
```
tom-un: The message in LoadFromDevServer is a wide string, but this one is narrow.  Perhaps make this std::wstring for consistency, and you won't need the explicit to_hstring below?   Of course then JSBundleNames would also have to be wstring

tom-un (approved):

windows/ReactTestApp/MainPage.cpp
tido64: Yes, I've changed them all to `std::wstring` in the latest iteration. Thanks for the review ❤️

PR #373, opened 2021-06-26
fix: warn and skip unsupported versions

tido64:
### Description

Warn and skip platforms that don't support target `react-native` version, e.g. `react-native-macos` when targeting `react-native` 0.64.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

```sh
yarn clean
yarn set-react-version 0.64
cd example
yarn
yarn configure-test-app --platforms macos
```

Since `react-native-macos` currently does not support 0.64, it should output:

```
[!] react-native-macos@0.64 cannot be added because it does not exist or is unsupported
```

…and exit without making any changes. If run with all platforms, it should exclude `react-native-macos`:

```
% yarn configure-test-app
[!] react-native-macos@0.64 cannot be added because it does not exist or is unsupported
[!] The following files will be overwritten:
     .watchmanconfig
     android/build.gradle
     android/gradle.properties
     android/gradlew
     android/gradlew.bat
     android/settings.gradle
     babel.config.js
     ios/Podfile
     metro.config.js
     react-native.config.js
[!] The following files will be removed:
     ios/Podfile.lock
[!] Destructive file operations are required.
Re-run with --force if you're fine with this.
```

### DIFF
--- a/example/yarn.lock
+++ b/example/yarn.lock
@@ -3615,6 +3615,7 @@ react-native-test-app@../:
     chalk "^4.1.0"
     prompts "^2.4.0"
     rimraf "^3.0.0"
+    semver "^7.3.5"
     yargs "^16.0.0"
 
 react-native-windows@^0.63.32:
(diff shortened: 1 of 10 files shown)


Saadnajmi:
Glad my struggle bus riding has led to improvements here 😅

tido64:
> Glad my struggle bus riding has led to improvements here 😅

Your struggles are always appreciated 😉

kelset (approved):

PR #387, opened 2021-07-02
fix(windows): `yarn windows` fails in generated projects

tido64:
### Description

Windows is not correctly configured, causing commands like `yarn windows` to fail.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [x] Windows

### Test plan

```
cd example
yarn
yarn windows
```

Tests were add. CI should pass.

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -22,11 +22,18 @@
   },
   "devDependencies": {
     "@babel/core": "^7.0.0",
+    "@types/jest": "^26.0.0",
+    "jest": "^27.0.0",
     "mkdirp": "^0.5.1",
     "react": "16.13.1",
     "react-native": "^0.63.4",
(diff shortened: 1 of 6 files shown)


example/package.json
```diff
@@ -22,11 +22,18 @@
   },
   "devDependencies": {
     "@babel/core": "^7.0.0",
+    "@types/jest": "^26.0.0",
```
kelset, line 25: if you use Jest 27 shouldn't the types also be in v27?

kelset: nvm they don't exist as of now: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/jest/index.d.ts

kelset (approved):

Issue #374, opened 2021-06-27
Add .gitignore with `init-test-app`


Saadnajmi:
### Proposal

Running `yarn init-test-app` didn't copy a .gitignore for me when I ran it. This meant that things like `dist/` and `example.xcworkspace` were showing up as available to be added to git. 

I ended up copying the one from this repo's example app for my local project. It would be nice if that's copied over.

### Additional context

Found while trying to move this module to RNTA: https://github.com/react-native-menu/menu

Shivakumar-SG:
Hey, @Saadnajmi 
I would like to take up this issue as my first contribution to OSS! Would you please guide me through it?

Saadnajmi:
Hello @Shivakumar-SG , I'm not an owner of this repo, so I haven't actually made a lot of changes here / know it's structure well. Someone else like @tido64 or @arazabishov might be better! However, I imagine you would probably want to start here https://github.com/microsoft/react-native-test-app/blob/trunk/scripts/configure.js
Probably add it to the common configuration? https://github.com/microsoft/react-native-test-app/blob/f87322599c8c0d4b0a510616ae1caf5b5f4978c4/scripts/configure.js#L287

tido64:
@Shivakumar-SG: Feel free to submit a PR 👍 

We probably want to copy both `/example/.gitignore` and `/example/windows/.gitignore`. The latter contains ignore rules for Visual Studio outputs. Once you've added them to the configure script, you'll also need to add them to `/.npmignore` to make sure they are included in the published packages.

Shivakumar-SG:
I thought of figuring out how the rest of the files were getting copied to the test app so that I can do the same for the mentioned .gitignore files but I could not able to trace them back. And if I try to remove one of the files getting copied by commenting it out or by removing, it still gets created in common files. Would you mind helping me with how these `babel.config.js`, `metro.config.js`, etc. are copied?

tido64:
@Shivakumar-SG: Can you please take a look at this PR and let me know if it has enough information for you: https://github.com/microsoft/react-native-test-app/pull/388

It is important for us that we have sufficient documentation in place so people can get up and running fast 😄

tido64:
Addressed by #389

PR #389, opened 2021-07-04
fix: .gitignore generated when running yarn init-test-app

Shivakumar-SG:
### Description

Added .gitignore files to `scripts/configure.js` and `/.npmignore`
 I actually merged both `example/.gitignore` and `example/windows/.gitignore` and added it into one. Which copies `.gitignore` files when running `yarn init-test-app`.
This solves issue #374 
Here is a sample screenshot :
Screenshot 2021-07-04 at 8 26 54 PM

### Platforms affected

- Android
- iOS
- macOS
- Windows

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -24,5 +24,5 @@ scripts/
 template-example/
 test/
 tsconfig.json
-example/.gitignore
-example/windows/.gitignore
+!example/.gitignore
+!example/windows/.gitignore
--- a/.npmignore
+++ b/.npmignore
@@ -24,3 +24,5 @@ scripts/
 template-example/
 test/
 tsconfig.json
+example/.gitignore
+example/windows/.gitignore
--- a/package.json
+++ b/package.json
@@ -95,6 +95,7 @@
     "react": "16.13.1",
     "react-native": "^0.63.4",
     "react-native-macos": "^0.63.33",
+    "react-native-test-app": "^0.6.10",
     "react-native-windows": "^0.63.32",
     "semantic-release": "^17.0.0",
     "suggestion-bot": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -95,7 +95,6 @@
     "react": "16.13.1",
     "react-native": "^0.63.4",
     "react-native-macos": "^0.63.33",
-    "react-native-test-app": "^0.6.10",
     "react-native-windows": "^0.63.32",
     "semantic-release": "^17.0.0",
     "suggestion-bot": "^1.0.0",
--- a/scripts/configure.js
+++ b/scripts/configure.js
@@ -323,6 +323,10 @@ const getConfig = (() => {
             "metro.config.js": {
               source: path.join(testAppPath, "example", "metro.config.js"),
             },
+            ".gitignore": fs
+              .readFileSync(path.join(process.cwd(), "example", ".gitignore"))
+               + "\n" + 
+               fs.readFileSync(path.join(process.cwd(), "example", "windows", ".gitignore")),
             "react-native.config.js": reactNativeConfig(params),
             ...(!init
               ? undefined
--- a/scripts/configure.js
+++ b/scripts/configure.js
@@ -327,10 +327,6 @@ const getConfig = (() => {
               .readFileSync(path.join(testAppPath, "example", ".gitignore"),{
                 encoding: "utf-8",
               }),
-            ".gitignorewinodws":
-               fs.readFileSync(path.join(testAppPath, "example", "windows", ".gitignore"),{
-                encoding: "utf-8",
-              }),
             "react-native.config.js": reactNativeConfig(params),
             ...(!init
               ? undefined
@@ -516,7 +512,12 @@ const getConfig = (() => {
           },
         },
         windows: {
-          files: {},
+          files: {
+            ".gitignorewinodws":
+               fs.readFileSync(path.join(testAppPath, "example", "windows", ".gitignore"),{
+                encoding: "utf-8",
+              }),
+          },
           oldFiles: [
             `${name}.sln`,
             `${name}.vcxproj`,
--- a/scripts/configure.js
+++ b/scripts/configure.js
@@ -513,7 +513,7 @@ const getConfig = (() => {
         },
         windows: {
           files: {
-            ".gitignorewinodws":
+            ".gitignorewindows":
                fs.readFileSync(path.join(testAppPath, "example", "windows", ".gitignore"),{
                 encoding: "utf-8",
               }),
--- a/scripts/configure.js
+++ b/scripts/configure.js
@@ -323,10 +323,9 @@ const getConfig = (() => {
             "metro.config.js": {
               source: path.join(testAppPath, "example", "metro.config.js"),
             },
-            ".gitignore": fs
-              .readFileSync(path.join(testAppPath, "example", ".gitignore"),{
-                encoding: "utf-8",
-              }),
+            ".gitignore": {
+              source: path.join(testAppPath, "example", ".gitignore"),
+            },
             "react-native.config.js": reactNativeConfig(params),
             ...(!init
               ? undefined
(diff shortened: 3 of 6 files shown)


.npmignore
```diff
@@ -24,3 +24,5 @@ scripts/
 template-example/
 test/
 tsconfig.json
+example/.gitignore
+example/windows/.gitignore
```
tido64: It looks like you're excluding `.gitignore` files here. Did you mean to add `!` in front?
```suggestion
!example/.gitignore
!example/windows/.gitignore

```

You should also add an newline to remove the warning:

image

package.json
```diff
@@ -95,6 +95,7 @@
     "react": "16.13.1",
     "react-native": "^0.63.4",
     "react-native-macos": "^0.63.33",
+    "react-native-test-app": "^0.6.10",
```
tido64: Why was this added? It shouldn't be necessary. Did you hit issues without it?

scripts/configure.js
```diff
@@ -323,6 +323,10 @@ const getConfig = (() => {
             "metro.config.js": {
               source: path.join(testAppPath, "example", "metro.config.js"),
             },
+            ".gitignore": fs
+              .readFileSync(path.join(process.cwd(), "example", ".gitignore"))
```
tido64: You shouldn't use current directory. We don't know where the user might be calling this script from. You should use `testAppPath` instead.

scripts/configure.js
```diff
@@ -323,6 +323,10 @@ const getConfig = (() => {
...
+               + "\n" +
```
tido64: I'm not sure you can just concatenate the two files. `windows/.gitignore` contains rules that may exclude names that are perfectly fine to have under the project folder. Can we instead copy the file under the Windows configuration section instead?

tido64 (changes requested):
Thanks for taking the time to do this. I've left some comments. Let me know if you have questions.

.npmignore
Shivakumar-SG: Yes, Sorry. I didn't know much about `.npmignore` and your feedback helped me. Thank you

package.json
Shivakumar-SG: No, I didn't hit any issues without it and it was auto-generated for me. So I thought I'd leave it that way. I will remove it.

Shivakumar-SG:
I have made the changes as suggested. Let me know if I have done things in the required way.

scripts/configure.js
```diff
@@ -323,6 +323,10 @@ const getConfig = (() => {
             "metro.config.js": {
               source: path.join(testAppPath, "example", "metro.config.js"),
             },
+            ".gitignore": fs
+              .readFileSync(path.join(testAppPath, "example", ".gitignore"),{
+                encoding: "utf-8",
+              }),
```
tido64: Since we're not making any changes to this file, we can just copy it:
```suggestion
            ".gitignore": {
               source: path.join(testAppPath, "example", ".gitignore"),
             },
```
Same below.

scripts/configure.js
```diff
@@ -508,7 +512,12 @@ const getConfig = (() => {
           },
         },
         windows: {
-          files: {},
+          files: {
+            ".gitignorewindows":
```
tido64: Why was this renamed? I don't think Git will pick it up if you do.
```suggestion
            ".gitignore":
```

yarn.lock
```diff
@@ -7875,6 +7875,17 @@ react-native-macos@^0.63.33:
     use-subscription "^1.0.0"
     whatwg-fetch "^3.0.0"
 
+react-native-test-app@^0.6.10:
```
tido64: Can you remove this as well?

tido64 (changes requested):

.npmignore
tido64: No worries. This file is a bit confusing. I forget to add things myself. 😄

scripts/configure.js
Shivakumar-SG: Oh, yea. This is all new to me so I just got confused and I feel stupid now.
Obviously, we have to have the name `.gitignore` in order to git to pick it up but it doesn't seem to copy contents from `example/windows/.gitignore`
What do you want me to do?

tido64: I made the changes locally and I can see the two new files just fine. How are you testing it? Can you push your changes so we can take a closer look?

Shivakumar-SG:
I did push the code. 
It seems to work fine when I select all platforms that we want to create a test app for. But If I select the platform as windows alone, it does not copy contents from `example/windows/.gitignore`

yarn.lock
```diff
@@ -7875,6 +7875,16 @@ react-native-macos@^0.63.33:
...
+  version "0.6.10"
+  resolved "https://registry.npmjs.org/react-native-test-app/-/react-native-test-app-0.6.10.tgz#426670261c43d17a4a1e74419cba6f5691dce88f"
+  integrity sha512-QnG+A9oaf20qRCsmnWh2WYfebmu/kn8DGN3OAUaNRpLaIh6j9/t5HIaT+6cn5bEmNpELCPaN1hfgyl2KmtE3RA==
+  dependencies:
+    chalk "^4.1.0"
+    prompts "^2.4.0"
+    rimraf "^3.0.0"
+    semver "^7.3.5"
+    yargs "^16.0.0"
+
```
tido64: Can you please revert this file, otherwise `yarn` will complain about an invalid lock file.

tido64:
> It seems to work fine when I select all platforms that we want to create a test app for. But If I select the platform as windows alone, it does not copy contents from `example/windows/.gitignore`

Right, it looks like the common config is overwriting the windows one here: https://github.com/microsoft/react-native-test-app/blob/3f03bea56bf7d72ae573e256d2870641e68b949f/scripts/configure.js#L616

We can probably reverse it but there are a couple of things I'd like to have a look at before we do that. In the meantime, I think we can merge this if you address the remaining comments. 👍

Shivakumar-SG:
> We can probably reverse it but there are a couple of things I'd like to have a look at before we do that. In the meantime, I think we can merge this if you address the remaining comments. 👍

I tried copying the `.gitignore` file into each platform's configuration section and in the window's section, I tried to concatenate both the `example/.gitignore` and `example/windows/.gitignore`, and its working
Here are the contents from the `.gitignore` created in this case:
Screenshot 2021-07-05 at 2 17 22 PM

And in other platforms too the `.gitignore` file is copied as required.
So is this right? or I just should 
>Can you please revert this file, otherwise yarn will complain about an invalid lock file.

address this issue and push the code?

tido64:
> So is this right? or I just should
> 
> > Can you please revert this file, otherwise yarn will complain about an invalid lock file.
> 
> address this issue and push the code?

I don't think that will be right. The rules in `.gitignore` is relative to the file path. If we move them into each platform, they will not apply the rules in the project root. Let's address the one issue and I can take a closer look later.

yarn.lock
tido64: Sorry, I meant can we undo all changes within this file. This section was added because you added `react-native-test-app` to `package.json` earlier.

yarn.lock
```diff
@@ -7875,6 +7875,7 @@ react-native-macos@^0.63.33:
     use-subscription "^1.0.0"
     whatwg-fetch "^3.0.0"
 
+
```
tido64: I think you missed a line when reverting this file 😄 
```suggestion
```

scripts/configure.js
```diff
@@ -508,7 +511,11 @@ const getConfig = (() => {
           },
         },
         windows: {
-          files: {},
+          files: {
+            ".gitignore": {
+               source: path.join(testAppPath, "example","windows", ".gitignore")
```
tido64: Build is currently failing due to a formatting issue. Can you please run `yarn format:js`? That should fix it up.

image

tido64 (changes requested):

tido64 (approved):
Looking good. Thanks for your hard work ❤️

Shivakumar-SG:
Thanks for giving me the opportunity to do it. This is my first contribution to Open Source and I felt good doing it. Thank you so much @tido64 for bearing with me throughout sir.

PR #413, opened 2021-07-15
docs: refresh README

tido64:
### Description

Refreshes `README`. Pads out `CONTRIBUTING`.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

There's nothing to test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,46 @@
 # React Native Test App
 
+[![Open in Visual Studio Code](https://open.vscode.dev/badges/open-in-vscode.svg)](https://open.vscode.dev/microsoft/react-native-test-app)
 ![build](https://github.com/microsoft/react-native-test-app/workflows/build/badge.svg)
-[![npm version](https://badgen.net/npm/v/react-native-test-app)](https://www.npmjs.com/package/react-native-test-app)
+[![npm version](https://img.shields.io/npm/v/react-native-test-app)](https://www.npmjs.com/package/react-native-test-app)
 
-React Native Test App provides test apps for all platforms as a package. Bring
-your own JS bundle, and let us manage the native bits.
+React Native Test App provides test apps for all platforms as a package. It
+handles the native bits for you so you can focus on what's important: Your
+product.
 
-This is a work in progress. You can read and discuss the RFC at
-[React Native: Discussions and Proposals](https://github.com/react-native-community/discussions-and-proposals/pull/204).
+## Motivation
 
-For more details, please refer to the
-[wiki](https://github.com/microsoft/react-native-test-app/wiki).
+Many of us have been there. We create a new React Native project with
+`react-native init`, write a few test screens for our library, and everything is
+seemingly fine and dandy. Until some time later, when a new version of React
+Native gets published. "Awesome!", you think, but when you try to upgrade to the
+latest version, something doesn't work. You need to look at
+[React Native Upgrade Helper](https://react-native-community.github.io/upgrade-helper/)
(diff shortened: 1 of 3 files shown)


arazabishov:
This is awesome! A suggestion that is outside the scope of this PR: we should have an example app showcasing the use of native modules and different types of native controllers (view controllers, fragments, activities, etc).

arazabishov (approved):

README.md
```diff
@@ -1,30 +1,60 @@
...
-
-This is a work in progress. You can read and discuss the RFC at
-[React Native: Discussions and Proposals](https://github.com/react-native-community/discussions-and-proposals/pull/204).
-
-For more details, please refer to the
-[wiki](https://github.com/microsoft/react-native-test-app/wiki).
+[![npm version](https://img.shields.io/npm/v/react-native-test-app)](https://www.npmjs.com/package/react-native-test-app)
+
+React Native Test App provides test apps for all platforms as a package. It
+handles the native bits for you so you can focus on what's important: Your
```
kelset:
```suggestion
handles the native bits for you so you can focus on what's important: your
```

README.md
```diff
@@ -1,30 +1,60 @@
...
+    compatibility or reproduce difficult bugs.
+  - So your next React Native upgrade is a lot more painless. Especially if you
+    have to repeat the same process in a lot of projects.
+- We want to make it easy to add support for and maintain additional platforms,
+  such as macOS or Windows, without requiring the domain knowledge to do so.
+- We want to give you a consistent developer experience across all the projects
+  you maintain.
+
+You can find the full design document in
+[the wiki](https://github.com/microsoft/react-native-test-app/wiki/Design).
```
kelset: I like this section, but I don't think many people want to see this as the first thing they read in the README; I would either move it lower down in the document or within the Wiki, and just have an extra line in the first paragraph like:

```md
If you want to learn more about [the motivation](link to wiki) or [the design](link to wiki) of this tool, you can refer to the wiki.
```

README.md
```diff
@@ -1,30 +1,60 @@
...
 
 ## Dependencies
```
kelset: I would move this section after the quick start: I'm afraid that this might lead to some confusion / some folks avoiding to use it because they read that they need to compile some stuff themselves.

But basically we can assume that all the developers here have an existing RN project.

tido64: Sounds good to me. I've deleted the section since we have a duplicate in the wiki. People can just go there if they really need it, but it shouldn't deviate from what they need for React Native in general.

README.md
```diff
@@ -1,54 +1,36 @@
...
-    [React Native for Windows documentation](https://microsoft.github.io/react-native-windows/docs/rnw-dependencies)
-  - [Google Chrome](https://www.google.com/chrome/) (optional, but recommended
-    for JS debugging)
-
-# Quick Start
+[![npm version](https://img.shields.io/npm/v/react-native-test-app)](https://www.npmjs.com/package/react-native-test-app)
+
+- [Quick Start](#quick-start)
+- [Configuring the Test App](#configuring-the-test-app)
+- [Known Issues](#known-issues)
```
kelset, line 9: super minor, feel free to ignore: this is not needed technically because GH has now added a table of content feature to READMEs

kelset (approved):
one minor thing but feel free to ignore and merge 👍

README.md
tido64: I feel it's less discoverable so I'm going to leave it there for now.

PR #429, opened 2021-07-28
chore: upgrade to Yarn 3

tido64:
### Description

Migrate to Yarn 3.0. The latest version allows us to set up a mini-monorepo so we don't need to keep two `yarn.lock`s (one in root and one under `/example`). I've set it up so that there is no hoisting between the two to mimic a "clean" app as much as possible. Yarn installs dependencies under `/example/node_modules` like before, but we no longer have to run `yarn link react-native-test-app` manually.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

Workflow should be simpler. For example, to boot the iOS app:

```
yarn
cd example
yarn ios
```


example/package.json
```diff
@@ -30,7 +29,7 @@
     "react": "16.13.1",
     "react-native": "^0.63.4",
     "react-native-macos": "^0.63.33",
-    "react-native-test-app": "../",
+    "react-native-test-app": "workspace:.",
```
kelset, line 32: iiinteresting - so basically now with Yarn 3 you need to define a local dependency through the workspace?

scripts/install-test-template.sh
```diff
@@ -34,19 +34,24 @@ done
...
 npm pack
 
-yarn ci
+yarn
 yarn react-native init-test-app --destination template-example --name TemplateExample --platform $platform
-cp .yarnrc-offline template-example/.yarnrc
 
 pushd template-example 1> /dev/null
 
+echo 'yarnPath: "../.yarn/releases/yarn-berry.cjs"' >> .yarnrc.yml
```
kelset: what's this yarn-berry file btw? isn't berry yarn 2?

example/package.json
tido64: Yeah, what forced the hand was the new checksum feature. It's different every time you ran `yarn install` inside `/example`. Converting to a workspace skips the check on local packages.

scripts/install-test-template.sh
tido64: It's apparently the "current" version:
```sh
% .yarn/releases/yarn-berry.cjs --version
3.0.0
```
With Yarn 3.0, it's recommended to check in a local copy. Your Yarn installation will automatically pick it up.

tido64:
~~Converting this back to a draft since we're blocked on https://github.com/dependabot/dependabot-core/issues/1297.~~

Edit: We've migrated to [Renovate](https://docs.renovatebot.com/) which supports Yarn 3 (#490).

scripts/install-test-template.sh
thoughtentity: The `yarn-berry.cjs` is the name `yarn` uses when you first run `yarn set version berry` to switch from `v1` to `v2` or `v3`. If you were to run `yarn set version latest` from a `v2` or `v3` install it'll switch over to using a versioned file name.

image

.github/workflows/build.yml
```diff
@@ -133,11 +129,10 @@ jobs:
           # https://github.com/facebook/react-native/issues/30036 and
           # https://github.com/microsoft/react-native-macos/issues/620 for more
           # details.
-          yarn set-react-version main
+          npm run set-react-version main
```
kelset, line 132: interesting, why is this going back to npm? Some yarn3 change that makes it more consistent to just go back to npm run?

example/ios/Podfile.lock
```diff
@@ -483,4 +483,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 192c89b7b9691a138c7ae651924b80cf83bccf1e
 
-COCOAPODS: 1.10.2
+COCOAPODS: 1.11.0
```
kelset: nit: cocoapods 1.11.2 is out

package.json
```diff
@@ -45,9 +45,7 @@
     "url": "https://github.com/microsoft/react-native-test-app.git"
   },
   "scripts": {
-    "ci": "yarn --use-yarnrc .yarnrc-offline --prefer-offline --frozen-lockfile --non-interactive",
-    "clean": "npx --quiet rimraf example/node_modules/react-native-test-app && git clean -dfqx",
-    "deduplicate": "npx yarn-deduplicate --strategy fewer",
```
kelset, line 50: we don't need dedup anymore because yarn 3 is better at it, right?

kelset (approved):
just a few Qs here and there, but honestly looks good

.github/workflows/build.yml
tido64: Yarn 3 doesn't allow you to `run` unless you've done a `yarn install` first. Something about security. I couldn't find an opt-out, so `npm run` it is 😄

package.json
tido64: Can't really say whether it's better, but the reason for removing this script is because it is now built in: https://yarnpkg.com/cli/dedupe

PR #443, opened 2021-08-10
fix: support for react-native 0.65

tido64:
### Description

Adds support for `react-native` 0.65.

### Platforms affected

- [x] Android
- [x] iOS
- [ ] macOS
- [ ] Windows

### Test plan

```sh
yarn set-react-version 0.65
cd example
rm ios/Podfile.lock  # not really necessary, just a precaution
yarn
yarn ios
```

| Home | App | App with Hermes |
|-|-|-|
| MicrosoftTeams-image (1) | MicrosoftTeams-image (2) | MicrosoftTeams-image (3) |


scripts/set-react-version.js
```diff
@@ -38,6 +38,14 @@ const profiles = {
     "react-native-macos": undefined,
     "react-native-windows": "^0.64.11",
   },
+  0.65: {
+    "hermes-engine": "~0.8.1",
+    react: "17.0.2",
+    "react-native": "^0.65.0-0",
+    "react-native-codegen": "0.0.7",
```
tido64, line 45: For whenever 0.65 is released:
```suggestion
    "react-native": "^0.65.0",
```

example/react-native.config.js
```diff
@@ -26,7 +26,30 @@ module.exports = {
       ),
     },
     ios: {
-      project: "ios/ReactTestApp-Dummy.xcodeproj",
+      project: (() => {
```
tido64: This needs to be ported to `scripts/configure.js` as well:
https://github.com/microsoft/react-native-test-app/blob/b72bb100d938d417fe7c0228693bcea29b84f37a/scripts/configure.js#L244-L246

scripts/set-react-version.js
tido64: Seems like `react-native-codegen` will be necessary for 0.65. 😞

example/App.js
```diff
@@ -52,21 +52,6 @@ const App = () => {
     backgroundColor: isDarkMode ? Colors.darker : Colors.lighter,
   };
 
-  const hermes = global.HermesInternal ? (
```
kelset, line 55: interesting, why did you remove the hermes section?

tido64: It was moved from the template to `NewAppScreen` in 0.64: https://github.com/facebook/react-native/commit/1c634a921887caf02f1b9fba6fd177c352f9ef78

kelset (approved):

PR #482, opened 2021-09-08
feat: populate home screen with AppRegistry entries

### LINKED ISSUE #189: Populate components list from AppRegistry if app.json is empty/missing
If the list of components in the manifest (`app.json`) is empty or missing, we should be able to retrieve the list of registered components from [`AppRegistry`](https://github.com/facebook/react-native/blob/21186034af265c093901c8b37497b600e97e4740/Libraries/ReactNative/AppRegistry.js#L147).

Remaining work:
- [ ] Android (#1127)
- [x] iOS (#482)
- [x] macOS (#482)
- [x] Windows (#482)

tido64:
### Description

When `components` is missing or empty in `app.json`, we attempt to populate the home screen with entries found in the `AppRegistry`.

Resolves #189.

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [x] Windows

### Test plan

1. Run `yarn` in example folder
   ```
   cd example
   yarn
   ```
2. Remove `components` from `app.json`:
   ```diff
   diff --git a/example/app.json b/example/app.json
   index 8463354..0d50753 100644
   --- a/example/app.json
   +++ b/example/app.json
   @@ -1,17 +1,6 @@
    {
      "name": "Example",
      "displayName": "Example",
   -  "components": [
   -    {
   -      "appKey": "Example",
   -      "displayName": "App"
   -    },
   -    {
   -      "appKey": "Example",
   -      "displayName": "App (modal)",
   -      "presentationStyle": "modal"
   -    }
   -  ],
      "resources": {
        "android": [
          "dist/res",
   ```
3. Build and run test app, e.g. for iOS:
   ```
   yarn ios
   
   # In a separate terminal:
   yarn start
   ```

### Remaining work

- [ ] Since this solution relies on JSI, we cannot reuse it on Android without requiring the user to install NDK. React Native currently does not have this requirement, and we don't want to add extra dependencies. A solution would be to build and ship AARs like react-native does, but this will have to be a separate PR.


windows/ReactTestApp/ReactInstance.cpp
```diff
@@ -57,12 +65,36 @@ namespace
...
-#endif
+#endif  // REACT_NATIVE_VERSION > 0 && REACT_NATIVE_VERSION < 6305
     struct DevSettings {
         REACT_INIT(Initialize)
-        void Initialize(winrt::Microsoft::ReactNative::ReactContext const &reactContext) noexcept
+        void Initialize(ReactContext const &reactContext) noexcept
         {
             context_ = reactContext;
+#if __has_include(<JSI/JsiApiContext.h>)
+            // There is currently no way to hook into when a JS bundle is done
```
acoates-ms: I think you want ReactInstanceSettings.InstanceLoaded?

windows/ReactTestApp/ReactInstance.cpp
```diff
@@ -57,12 +65,36 @@ namespace
...
+            // loading. However, it seems that by posting a callback, we can
+            // defer fetching app keys until everything is done executing.
+            reactContext.JSDispatcher().Post([reactContext]() noexcept {
+                auto &runtime =
```
acoates-ms: What happens if you get here when webdebugging?

tido64: This is a good call. The test app is broken due to an unrelated issue fixed in #493.

With the fix applied, what happens is that we return early with an empty array here:
https://github.com/microsoft/react-native-test-app/blob/09fca861f0943e82b7890aad88ad4204150364b0/common/AppRegistry.cpp#L22-L25

I'm hesitant to implement a fallback for this, especially when web debugging might go away soon. Unless you have any suggestions, I'm going to leave this as is.

arazabishov:
Could we send the list of components over the bridge instead? (to include android)

tido64:
> Could we send the list of components over the bridge instead? (to include android)

We could, but that requires changes to the JS. I'm not sure how we can achieve that without requiring users to add an extra import or similar.

macos/ReactTestApp/AppDelegate.swift
```diff
@@ -41,24 +45,34 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
...
+                }
             )
-            item.tag = index
-            item.keyEquivalentModifierMask = [.shift, .command]
-            item.isEnabled = false
-            item.representedObject = component
         }
 
-        let components = manifest.components
+        onComponentsRegistered(components, enable: false)
```
tom-un, line 70: If there is a list of componetns in the manifest, wouldn't you want to call this with enable: true?

tido64: The property is for deciding whether the menu item should be enabled. Since the bridge hasn't been initialized yet, we should prevent them from being clicked. They get enabled when it is:

https://github.com/microsoft/react-native-test-app/blob/fae8b9fd5078e6c0d2d2cf0b6e94a29f58e1e712/macos/ReactTestApp/AppDelegate.swift#L83

tom-un (approved):

PR #495, opened 2021-09-15
fix(android): remove workaround for trove4j

tido64:
### Description

`org.jetbrains.trove4j:trove4j` has been uploaded to Maven Central: https://search.maven.org/artifact/org.jetbrains.trove4j/trove4j/20160824/jar

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

Green Android build.


tido64:
Just realised that this is a breaking change since we're removing a file that may be in use. I'm moving this to a different PR.

PR #508, opened 2021-09-20
fix(apple): reduce build noise from Flipper and dependencies

tido64:
### Description

Warnings were disabled for Flipper and its dependencies because there are too many, making it difficult to see real issues.

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [ ] Windows

### Test plan

Green CI.

### DIFF
--- a/ios/ReactTestApp.xcodeproj/project.pbxproj
+++ b/ios/ReactTestApp.xcodeproj/project.pbxproj
@@ -223,8 +223,8 @@
 		19ECD0CA232ED425003D8557 /* Project object */ = {
 			isa = PBXProject;
 			attributes = {
-				LastSwiftUpdateCheck = 1250;
-				LastUpgradeCheck = 1250;
+				LastSwiftUpdateCheck = 1300;
+				LastUpgradeCheck = 1300;
 				ORGANIZATIONNAME = Microsoft;
 				TargetAttributes = {
 					19ECD0D1232ED425003D8557 = {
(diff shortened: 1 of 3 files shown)


ios/ReactTestApp.xcodeproj/project.pbxproj
```diff
@@ -223,8 +223,8 @@
 		19ECD0CA232ED425003D8557 /* Project object */ = {
 			isa = PBXProject;
 			attributes = {
-				LastSwiftUpdateCheck = 1250;
-				LastUpgradeCheck = 1250;
+				LastSwiftUpdateCheck = 1300;
+				LastUpgradeCheck = 1300;
```
kelset, line 227: probably just a nit, but might be worth waiting for this change to land for when Xcode 13.0 comes out? Or write a note somewhere about the fact that we're already using Xcode 13 here?

Or just do nothing, I guess it's mostly just you using this repo anyway 😅

kelset (approved):

ios/ReactTestApp.xcodeproj/project.pbxproj
tido64: These two lines don't really do anything other than letting Xcode know that we've done a code base "upgrade". There were no changes between 12 and 13 in RNTA, so these have no other effect than not prompting users to upgrade when they open a project with Xcode 13.

PR #529, opened 2021-09-27
feat: add support for custom bundle root

### LINKED ISSUE #525: Allow custom bundle roots
### Proposal

React Native Test App currently only recognises two bundle roots, `index` and `main` (see [How the Test App Finds the JS Bundle](https://github.com/microsoft/react-native-test-app/wiki#how-the-test-app-finds-the-js-bundle)). There should be a way to change the bundle root.

### Alternatives

n/a

### Implementation Details

A new, optional property in the manifest:

```json
{
  "bundleRoot": "myRoot"
}
```

When it is unset, the test app should use `index` and `main` as before. This behaviour is for convenience, and ensures backwards compatibility.

When it is set, the test app will use the value instead. For instance, if it is set to `myRoot`, the test app will look for the following files:
- `myRoot.[platform].jsbundle`
- `myRoot.[platform].bundle`
- `myRoot.native.jsbundle`
- `myRoot.native.bundle`
- `myRoot.jsbundle`
- `myRoot.bundle`

### Additional Context

_No response_

### Code of Conduct

tido64:
### Description

Adds support for custom bundle root.

Resolves #525.

### Platforms affected

- [x] Android
- [x] iOS
- [x] macOS
- [x] Windows

### Test plan

1. When `bundleRoot` is unset, test app should use `index` and `main` as bundle root.
2. When `bundleRoot` is set, test app should only use the set value as bundle root.

### DIFF
--- a/android/app/src/main/java/com/microsoft/reacttestapp/TestApp.kt
+++ b/android/app/src/main/java/com/microsoft/reacttestapp/TestApp.kt
@@ -13,17 +13,19 @@ val Context.testApp: TestApp
     get() = applicationContext as TestApp
 
 class TestApp : Application(), ReactApplication {
+    private lateinit var manifestProviderInternal: ManifestProvider
     private lateinit var reactNativeBundleNameProvider: ReactBundleNameProvider
     private lateinit var reactNativeHostInternal: TestAppReactNativeHost
-    private lateinit var manifestProviderInternal: ManifestProvider
 
     override fun onCreate() {
         super.onCreate()
 
-        reactNativeBundleNameProvider = ReactBundleNameProvider(this)
+        manifestProviderInternal = ManifestProvider.create(this)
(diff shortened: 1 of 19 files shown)


android/app/src/main/java/com/microsoft/reacttestapp/TestApp.kt
```diff
@@ -13,17 +13,19 @@ val Context.testApp: TestApp
     get() = applicationContext as TestApp
 
 class TestApp : Application(), ReactApplication {
+    private lateinit var manifestProviderInternal: ManifestProvider
```
arazabishov, line 16: Q: what do we need the internal postfix for? :)

arazabishov: Ah, I remember now. I think there was a name with variables in ReactApplication.

arazabishov (approved):

PR #553, opened 2021-10-13
fix(macos): Fixes to macOS test app to suport MSAL authentication

tom-un:
### Description

The macOS test app had two issues the prevented test apps from using the MSAL auth framework:
1) The macOS `AppDelegate`'s  `applicationDidFinishLaunching` method was not firing the `ReactTestAppDidInitialize` notification as its iOS counterpart does.  This notification is used in internal react-native-test-app support packages that provide MSAL auth.
2) The macOS MSAL framework depends on group keychain access to `com.microsoft.identity.universalstorage`.  This is similar to the iOS MSAL framework that depends on group keychain access to `com.microsoft.adalcache`.  See [MSIDKeychainTokenCache.m](https://github.com/AzureAD/microsoft-authentication-library-common-for-objc/blob/4f1855df645ac58cfa39dbe0fb8ba796c5c6ec34/IdentityCore/src/cache/MSIDKeychainTokenCache.m#L44).
  - As a consequence of adding this entitlement, the macOS test app must be signed.  So the macOS pbxproj was updated to sign similar to its iOS counterpart.

### Platforms affected

- [ ] Android
- [ ] iOS
- [x] macOS
- [ ] Windows

### Test plan

- Verified that `yarn macos` builds and runs in the `example` folder.
- Ingested into local macOS test app that authenticates using MSAL framework and verified that it can successful auth.

### DIFF
--- a/macos/ReactTestApp.xcodeproj/project.pbxproj
+++ b/macos/ReactTestApp.xcodeproj/project.pbxproj
@@ -8,14 +8,14 @@
 
 /* Begin PBXBuildFile section */
 		193EF063247A736200BE8C79 /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 193EF062247A736200BE8C79 /* AppDelegate.swift */; };
-		193EF08F247A799D00BE8C79 /* Manifest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 193EF08E247A799D00BE8C79 /* Manifest.swift */; };
-		193EF098247B130700BE8C79 /* ReactInstance.swift in Sources */ = {isa = PBXBuildFile; fileRef = 193EF097247B130700BE8C79 /* ReactInstance.swift */; };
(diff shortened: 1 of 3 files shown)


macos/ReactTestApp.xcodeproj/project.pbxproj
```diff
@@ -379,9 +379,10 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = ReactTestApp/ReactTestApp.entitlements;
-				CODE_SIGN_IDENTITY = "-";
+				CODE_SIGN_IDENTITY = "Apple Development";
 				CODE_SIGN_STYLE = Automatic;
 				COMBINE_HIDPI_IMAGES = YES;
+				DEVELOPMENT_TEAM = UBF8T346G9;
```
tido64, line 385: Won't this fail for other people when Xcode can't find the team?

tom-un: I don't think so because "Automatic signing" is on.  The same development team string is in the iOS project.pbxproj

tido64 (approved):
Thanks for fixing this ❤️

PR #562, opened 2021-10-22
test(android): run unit tests locally to speed up Android CI

tido64:
### Description

Trying out Ubuntu VMs to see if they are faster or more reliable than the macOS ones.

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

CI should be green (and faster).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -196,7 +196,7 @@ jobs:
     name: "Android"
     strategy:
       matrix:
-        os: [macos-11, windows-latest]
+        os: [ubuntu-latest, windows-latest]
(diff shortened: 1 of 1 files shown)


kelset:
so I guess they weren't faster? :D

tido64:
> so I guess they weren't faster? :D

It didn't. Intel HAXM is only installed on macOS VMs. But if you look at how many tests we have (one, we have 1), it shouldn't take 7 minutes (!) to execute. I'm currently looking into Robolectric to run it locally instead of inside an emulator.

@arazabishov: Did we ever consider Robolectric?

Update: #563

PR #580, opened 2021-11-05
chore: disable scripts when installing the project

tido64:
### Description

Don't run scripts when installing the project.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

CI should pass.

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,3 +1,4 @@
+enableScripts: false
 enableTelemetry: false
 logFilters:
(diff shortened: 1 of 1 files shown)


Saadnajmi:
Is this as a response to recent malware in npm packages?

tido64:
> Is this as a response to recent malware in npm packages?

Yeah, pretty much. It doesn't look like it is used a lot, so we might as well disable it.

PR #607, opened 2021-11-26
ci(nightlies): fetch latest react-native-windows canary from NuGet

tido64:
### Description

Nightly builds with react-native-windows@canary often fail due to the NuGet feed being out of sync with npm.

- When the latest nightly was built, .422 was the latest canary and could not be found on the NuGet feed: https://github.com/microsoft/react-native-test-app/runs/4330499508?check_suite_focus=true
- When I reran the job, the latest became .423: https://github.com/microsoft/react-native-test-app/runs/4334664457?check_suite_focus=true
- With the fix, we are building .422 instead: https://github.com/microsoft/react-native-test-app/runs/4335061574?check_suite_focus=true

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [x] Windows

### Test plan

This "nightly" build should be green: https://github.com/microsoft/react-native-test-app/runs/4335061574?check_suite_focus=true

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -400,13 +400,13 @@ jobs:
           path: .yarn/cache
           key: ${{ hashFiles('yarn.lock') }}
       - name: Set up react-native@canary
-        if: ${{ github.event_name != 'schedule' }}
+        if: ${{ github.event_name == 'schedule' }}
         run: |
           npm run set-react-version canary-windows
           yarn --no-immutable
         shell: bash
       - name: Install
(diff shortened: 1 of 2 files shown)


NickGerleman (approved):
This is a reminder for us to fix the whole issue of allowing publish of NPM packages when NuGet publish will fail or come half an hour later. cc @dannyvv @jonthysell 

Not sure when it will be able to be fixed though (there's a large architecture change we've wanted to make that gets us there), so this change seems righteous.

PR #630, opened 2021-12-04
ci: fix conditional expressions in composite actions

tido64:
### Description

It doesn't look like booleans are being passed to composite actions as is, but are serialized or something.

Nightly builds are failing because yarn action passes `immutable` as a string instead of a boolean: https://github.com/microsoft/react-native-test-app/runs/4415712667?check_suite_focus=true

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

CI should pass.


tido64:
image
Comparing as strings works.

PR #631, opened 2021-12-04
ci: hash `yarn.lock` ourselves for better consistency

tido64:
### Description

GitHub's `hashFiles()` function creates platform unique keys. Yarn's cache only contains tarballs and should be shareable across all platforms.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

CI should pass.

### DIFF
--- a/.github/actions/init-test-app/action.yml
+++ b/.github/actions/init-test-app/action.yml
@@ -7,11 +7,15 @@ inputs:
 runs:
   using: composite
   steps:
+    - name: Generate cache key
+      id: cache-key-generator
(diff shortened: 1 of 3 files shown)


tido64:
Caches aren't shared between platforms:

Ubuntu:

image

Windows:

image

😕

tido64:
But we can hopefully make the hash more consistent between trunk and PR builds (`yarn.lock` hasn't changed between these two builds):

trunk:

image

PR:

image

tido64:
image

Aaaand it looks like `trunk` has its own cache… 😕

PR #646, opened 2021-12-10
chore(windows): script for renewing the self-signed test certificate

tido64:
### Description

Automate renewal of the self-signed test certificate for ReactTestApp on Windows.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [x] Windows

### Test plan

~~The renewal workflow has been configured to trigger on this PR. A new PR should be created with new and valid test certificate.~~

### DIFF
--- a/.github/workflows/renew-windows-certificate.yml
+++ b/.github/workflows/renew-windows-certificate.yml
@@ -1,8 +1,8 @@
 name: Reminder to Renew Self-signed Test Certificate
 on:
   schedule:
-    # 04:05 on Monday in January and June
-    - cron: 5 4 * 1,6 1
+    # 04:05 on Monday in January and July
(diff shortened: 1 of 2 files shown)


.github/workflows/renew-windows-certificate.yml
```diff
@@ -1,23 +1,26 @@
-name: Reminder to Renew Self-signed Test Certificate
+name: Renew Self-signed Test Certificate
 on:
+  pull_request:
```
tido64: Remove this trigger when the workflow works.

tido64:
image

The script works locally, but when run on CI, the new certificate is missing a private key. For now, we'll just check in the script and update the reminder message to use the script.

PR #677, opened 2022-01-02
fix(ios): replace QRCodeReader with custom implementation

tido64:
### Description

Replaces [QRCodeReader](https://github.com/YannickL/QRCodeReader.swift) with our own implementation.

### Platforms affected

- [ ] Android
- [x] iOS
- [ ] macOS
- [ ] Windows

### Test plan

1. Build and run iOS test app
2. Shake the device and tap "Scan QR Code"
3. Verify that rotation works
4. Verify that QR code scanning still works:
    image
    `http://localhost:8081/index.bundle`


tom-un:
Was  QRCodeReader just become too old and unmaintained?

tom-un (approved):

tido64:
> Was QRCodeReader just become too old and unmaintained?

Pretty much. There are several unmerged PRs and Xcode has started complaining about the Swift code. There isn't much code to enable this feature anyway. This way we have better control and one less dependency to worry about.

PR #692, opened 2022-01-06
feat(android): Add ability to configure key store

### LINKED ISSUE #691: Android: Add ability to configure key store
### Proposal

The ability to [configure the key store](https://developer.android.com/studio/publish/app-signing) is required for enabling certain features such as MSAL's [auth broker](https://github.com/microsoft/rnx-kit/issues/900).

### Alternatives

n/a

### Implementation Details

We need to add `signingConfigs` property to the Android specific section in the [app manifest](https://github.com/microsoft/react-native-test-app/wiki/Manifest-(app.json)#android). Where possible, this new property should closely match the [DSL](https://developer.android.com/reference/tools/gradle-api/4.2/com/android/build/api/dsl/SigningConfig). I'm thinking the following is a good start:

```typescript
{
  "android": {
    "signingConfigs": {
      "config": {
        "keyAlias": string | undefined;
        "keyPassword": string | undefined;
        "storeFile": string | undefined;
        "storePassword": string | undefined;
      },
      "debug": {
        "keyAlias": string | undefined;
        "keyPassword": string | undefined;
        "storeFile": string | undefined;
        "storePassword": string | undefined;
      },
      "release": {
        "keyAlias": string | undefined;
        "keyPassword": string | undefined;
        "storeFile": string | undefined;
        "storePassword": string | undefined;
      }
    }
  }
}
```

#### What needs to be updated

Somewhere in the `android` section, we need to parse the manifest and call the appropriate functions:

https://github.com/microsoft/react-native-test-app/blob/43e2c9294a05a1d1b108c9d4ef76c4d590d1c7a0/android/app/build.gradle#L59-L109

#### How to teach this

- Update the [JSON schema](https://github.com/microsoft/react-native-test-app/blob/trunk/schema.json)
- Update [Android documentation](https://github.com/microsoft/react-native-test-app/wiki/Android-Specifics)
- Update [Manifest documentation](https://github.com/microsoft/react-native-test-app/wiki/Manifest-(app.json)#android)

### Additional Context

_No response_

### Code of Conduct

sveinar:
The ability to configure the key store is required for enabling certain features such as MSAL's auth broker.

### Description

We need to add signingConfigs property to the Android specific section in the app manifest.

I tried as much to match the DSL of the app manifest itself. 
In the code, I have 'landed' on a fixed set of flavors, debug and release.

The trouble to make this generic on the fly in the 'android' section of the gradle is that the signingConfigs are ReadOnly object when first created. If I tried to use closures inside the 'android' section with .each { flavor -> ... } then there was a mismatch with the closure and the DSL, even if I tried to use a proper Java class for the SigningConfig. Alas, two flavors for now.

Resolves #691.

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

Tested with overwriting the 'debug' flavor using a debug.keystore. 
Tested with a release keystore.
Tested without any signingConfig in the app.json.
Tested with signingConfig with a 
  - missing keystore
  - keystore just as '..' or any other folder
  - keystore empty

### TODO

- [x] Update the Android documentation
- [x] Update the Manifest documentation

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -56,7 +56,31 @@ project.ext.react = [
     enableHermes : true,
 ]
 
+def signingDebug = getSigningConfigs()['debug']
+def signingRelease = getSigningConfigs()['release']
+
 android {
+    if (signingDebug || signingRelease) {
+        signingConfigs {
+            if (signingDebug) {
+                debug {
+                    keyAlias signingDebug['keyAlias']
+                    keyPassword signingDebug['keyPassword']
+                    storeFile signingDebug['storeFile']
+                    storePassword signingDebug['storePassword']
+                }
+            }
+            if (signingRelease) {
+                release {
+                    keyAlias signingRelease['keyAlias']
+                    keyPassword signingRelease['keyPassword']
+                    storeFile signingRelease['storeFile']
+                    storePassword signingRelease['storePassword']
+                }
+            }
+        }
+    }
+
     compileSdkVersion project.ext.compileSdkVersion
     buildToolsVersion project.ext.buildToolsVersion
 
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -56,26 +56,27 @@ project.ext.react = [
     enableHermes : true,
 ]
 
-def signingDebug = getSigningConfigs()['debug']
-def signingRelease = getSigningConfigs()['release']
+project.ext.signingConfigs = getSigningConfigs()
 
 android {
+    def signingDebug = project.ext.signingConfigs["debug"]
+    def signingRelease = project.ext.signingConfigs["release"]
     if (signingDebug || signingRelease) {
         signingConfigs {
             if (signingDebug) {
                 debug {
-                    keyAlias signingDebug['keyAlias']
-                    keyPassword signingDebug['keyPassword']
-                    storeFile signingDebug['storeFile']
-                    storePassword signingDebug['storePassword']
+                    keyAlias signingDebug["keyAlias"]
+                    keyPassword signingDebug["keyPassword"]
+                    storeFile signingDebug["storeFile"]
+                    storePassword signingDebug["storePassword"]
                 }
             }
             if (signingRelease) {
                 release {
-                    keyAlias signingRelease['keyAlias']
-                    keyPassword signingRelease['keyPassword']
-                    storeFile signingRelease['storeFile']
-                    storePassword signingRelease['storePassword']
+                    keyAlias signingRelease["keyAlias"]
+                    keyPassword signingRelease["keyPassword"]
+                    storeFile signingRelease["storeFile"]
+                    storePassword signingRelease["storePassword"]
                 }
             }
         }
--- a/android/test-app-util.gradle
+++ b/android/test-app-util.gradle
@@ -145,3 +145,41 @@ ext.getReactNativeVersionNumber = { baseDir ->
     def (major, minor, patch) = manifest["version"].findAll(/\d+/)
     return (major as int) * 10000 + (minor as int) * 100 + (patch as int)
 }
+
+ext.getSigningConfigs = { baseDir ->
+    def safeSetMap = { varName, map, prop, defaultVal ->
+        map[varName] = prop.containsKey(varName) ? prop.get(varName) : defaultVal
+    }
+
+    def definedConfigs = new LinkedHashMap<String, Object>()
+    def manifestFile = findFile('app.json')
+    if (manifestFile != null) {
+        def manifest = new JsonSlurper().parseText(manifestFile.text)
+
+        if (!manifest['android']) {
+            return definedConfigs
+        }
+
+        def signingConfigs = manifest['android']['signingConfigs']
+        if (signingConfigs) {
+            signingConfigs.each { config ->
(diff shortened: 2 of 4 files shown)


android/app/build.gradle
```diff
@@ -56,7 +56,31 @@ project.ext.react = [
     enableHermes : true,
 ]
 
+def signingDebug = getSigningConfigs()['debug']
+def signingRelease = getSigningConfigs()['release']
+
 android {
+    if (signingDebug || signingRelease) {
+        signingConfigs {
```
tido64: Maybe something like this to reduce the number of times we parse `app.json`?
```suggestion
project.ext.signingConfigs = getSigningConfigs()

android {
    def signingDebug = project.ext.signingConfigs['debug']
    def signingRelease = project.ext.signingConfigs['release']
    if (signingDebug || signingRelease) {
        signingConfigs {
```

schema.json
```diff
@@ -83,6 +83,38 @@
...
+            "keyPassword": {
+              "type": "string"
+            },
+            "storeFile": {
+              "type": "string"
+            },
+            "storePassword": {
+              "type": "string"
+            }
+          }
```
tido64: I think we need to wrap the properties inside a `properties` property 😄 
```suggestion
          "properties": {
            "debug": {
              "type": "object",
              "properties": {
                "keyAlias": {
                  "type": "string"
                },
                "keyPassword": {
                  "type": "string"
                },
                "storeFile": {
                  "type": "string"
                },
                "storePassword": {
                  "type": "string"
                }
              },
              "required": ["storeFile"]
            },
            "release": {
              "type": "object",
              "properties": {
                "keyAlias": {
                  "type": "string"
                },
                "keyPassword": {
                  "type": "string"
                },
                "storeFile": {
                  "type": "string"
                },
                "storePassword": {
                  "type": "string"
                }
              },
              "required": ["storeFile"]
            }
          }
```
It would also be great to add the `description` field to each of these.

You can also use `yarn codegen` to verify this file.

android/test-app-util.gradle
```diff
@@ -145,3 +145,41 @@ ext.getReactNativeVersionNumber = { baseDir ->
...
+            return definedConfigs
+        }
+
+        def signingConfigs = manifest['android']['signingConfigs']
+        if (signingConfigs) {
+            signingConfigs.each { config ->
+                def configName = config.key
+                def props = config.value
+                def pathStoreFile = props.containsKey('storeFile') ?
+                    Paths.get(rootDir.toString(), props.get('storeFile')).normalize().toAbsolutePath() : null
```
tido64: Nit: Formatting:
```suggestion
                def pathStoreFile = props.containsKey('storeFile')
                    ? Paths.get(rootDir.toString(), props.get('storeFile')).normalize().toAbsolutePath()
                    : null
```

android/app/build.gradle
```diff
@@ -56,7 +56,31 @@ project.ext.react = [
     enableHermes : true,
 ]
 
+def signingDebug = getSigningConfigs()['debug']
```
tido64: Can we replace all the single-quotes with doubles? When we migrate to the Kotlin DSL later, it will require all strings to be double-quoted.

tido64 (commented):
This looks great! Just a few nitpicks and it will be good to go.

android/test-app-util.gradle
```diff
@@ -145,3 +145,41 @@ ext.getReactNativeVersionNumber = { baseDir ->
     def (major, minor, patch) = manifest["version"].findAll(/\d+/)
     return (major as int) * 10000 + (minor as int) * 100 + (patch as int)
 }
+
+ext.getSigningConfigs = { baseDir ->
```
tido64, line 149: Can we add tests for this new function? See https://github.com/microsoft/react-native-test-app/blob/trunk/test/android-test-app/test-app-util.test.js

sveinar: Added a simple test :) Thanks!

schema.json
sveinar: Good point!

android/app/build.gradle
sveinar: thanks! done

android/test-app-util.gradle
sveinar: nice!

android/test-app-util.gradle
```diff
@@ -145,3 +145,42 @@ ext.getReactNativeVersionNumber = { baseDir ->
...
+            return definedConfigs
+        }
+
+        def signingConfigs = manifest["android"]["signingConfigs"]
+        if (signingConfigs) {
+            signingConfigs.each { config ->
+                def configName = config.key
+                def props = config.value
+                def pathStoreFile = props.containsKey("storeFile")
+                    ? Paths.get(rootDir.toString(), props.get("storeFile")).normalize().toAbsolutePath()
```
tido64: Should this be relative to `app.json`? `rootDir` depends on where the `build.gradle` file lives, iirc.

sveinar: Great find. Yes, you are correct. Pushed a change for that to use the manifest's location as a reference.

tido64 (approved):
Thanks again for this ❤️

PR #704, opened 2022-01-18
fix(macos): prepare support for react-native-macos 0.66

tido64:
### Description

Make sure to call `__apply_Xcode_12_5_M1_post_install_workaround` if it is defined.

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [ ] Windows

### Test plan

```
npm run set-react-version canary-macos
yarn
cd example
../scripts/xcodebuild.sh macos/Example.xcworkspace build

# Also make sure iOS builds
../scripts/xcodebuild.sh ios/Example.xcworkspace build
```

### DIFF
--- a/scripts/set-react-version.js
+++ b/scripts/set-react-version.js
@@ -15,7 +15,7 @@ const os = require("os");
 
 const VALID_TAGS = ["canary-macos", "canary-windows", "main", "nightly"];
 const REACT_NATIVE_VERSIONS = {
-  "canary-macos": "^0.64",
+  "canary-macos": "^0.66",
   "canary-windows": "^0.67.0-0",
 };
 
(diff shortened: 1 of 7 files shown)


scripts/set-react-version.js
```diff
@@ -15,7 +15,7 @@ const os = require("os");
 
 const VALID_TAGS = ["canary-macos", "canary-windows", "main", "nightly"];
 const REACT_NATIVE_VERSIONS = {
-  "canary-macos": "^0.64",
+  "canary-macos": "^0.66",
```
kelset, line 18: just a simple question for me to learn - here you can put 0.66 already even if it's not out yet?

kelset: (as far as I can see, the canary macos on npm have the `0.0.0-hash` format)

tido64: This is for pointing at the correct version of react-native to install with react-native-macos. The latter doesn't specify which version to use because it is self-contained, but I think cli complains if it cannot find react-native.

kelset (approved):

PR #705, opened 2022-01-20
fix: `validate-manifest` breaks on Node 12

tido64:
### Description

`validate-manifest` breaks on Node 12 due to use of optional chaining.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

CI should pass.

### DIFF
--- a/scripts/validate-manifest.js
+++ b/scripts/validate-manifest.js
@@ -47,12 +47,15 @@ function validateManifest(manifestPath = findAppManifest()) {
     console.error(
       `${manifestPath}: error: ${APP_JSON} is not a valid app manifest`
     );
-    validate.errors?.map(({ instancePath, message }) =>
-      console.error(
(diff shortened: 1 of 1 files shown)


kelset:
Do we still need to support Node 12? Current LTS is 16 https://nodejs.org/en/

tido64:
> Do we still need to support Node 12? Current LTS is 16 https://nodejs.org/en/

Unfortunately, react-native-macos is still using Node 12 on their CI.

kelset (approved):

Saadnajmi:
> > Do we still need to support Node 12? Current LTS is 16 https://nodejs.org/en/
> 
> Unfortunately, react-native-macos is still using Node 12 on their CI.

Probably a good action item for us to update our CI too. Especially since it's out of LTS.

PR #741, opened 2022-02-03
chore(ios): add hardening on build to catch potential security issues

crazyfraggle:
### Description

Adding CFLAGS and LDFLAGS for somewhat hardened builds. This is a requirement internally in Microsoft for apps that are built in CI. 

This PR adds flags for iOS. Similar fixes will be needed for MacOS and Android as well.

### Platforms affected

- [ ] Android
- [X] iOS
- [ ] macOS
- [ ] Windows

### Test plan

The hardening should catch potential future issues in test applications and included libraries. Manual testing has so far not revealed any issues though.


ios/ReactTestApp/ReactTestApp.common.xcconfig
```diff
@@ -16,3 +16,6 @@ MTL_FAST_MATH = YES
 SDKROOT = iphoneos
 SWIFT_TREAT_WARNINGS_AS_ERRORS = YES
 WARNING_CFLAGS = -Wall
+OTHER_CFLAGS = $(inherited) -fstack-protector-strong -fPIE -Wformat -Wformat-security -Werror=format-security
+OTHER_LDFLAGS = $(inherited) -fstack-protector-strong -fPIE -pie -Wformat -Wformat-security -Werror=format-security
```
tido64: `-W*` shouldn't be necessary since we have `WARNING_CFLAGS = -Wall` and `GCC_TREAT_WARNINGS_AS_ERRORS = YES`:
```suggestion
OTHER_CFLAGS = $(inherited) -fstack-protector-strong -fPIE
OTHER_LDFLAGS = $(inherited) -fstack-protector-strong -fPIE -pie
```

Also, can you sort the list? 😄

tido64: Is `-fPIE` still necessary when linking? I thought `-pie` was enough?

tido64: Curiously, enabling `Position-Dependent Executable`:

image

Adds `LD_NO_PIE = YES` to `project.pbxproj` ¯\\_(ツ)_/¯

crazyfraggle: Yeah, -Wall enables -Wmost which enables -Wformat which enables -Wformat-securtity. (Most of which are enabled by default anyway...) So yeah, turns out to be redundant. :/

crazyfraggle: I guess we shouldn't select that then. And that implies to me that PIE is default, right?

tido64: Yes, it implies that PIE is default so it looks like `-pie` is redundant. I also checked the output, it adds `-no-pie` to disable it.

tido64: Which leaves `-fstack-protector-strong` as the only new flag here. I'm not sure it is necessary when linking though. Is that something you can verify?

crazyfraggle: I can check if it's required on linking.

crazyfraggle: Seems to be recommended to also include -fstack-protector-strong on the ldflags as well.

tido64: Thanks for checking 😄

tido64 (approved):
Thanks for doing this! Will you also look into Android?

PR #754, opened 2022-02-07
ci: cache compiler outputs to speed up builds

tido64:
### Description

iOS/macOS builds can exceed 20 minutes nowadays. Experimenting with different solutions to see if we can speed them up. Things that I've attempted but didn't work out:

- Caching DerivedData
  - Xcode rebuilds everything anyway, despite touching all files to update the file modified times and telling `xcodebuild` to ignore inode changes
  - The DerivedData folder was also very big
- [PodBuilder](https://github.com/Subito-it/PodBuilder)
  - Builds and caches Pods only
  - Requires extra build steps, and modifying `Podfile`
- [CocoaPods binary cache](https://github.com/grab/cocoapods-binary-cache)
  - Builds and caches Pods only
  - Requires extra build steps, and modifying `Podfile` and our react-native profiles
- [CocoaPods Binary](https://github.com/leavez/cocoapods-binary)
  - Builds and caches Pods only
  - Requires extra build steps, and modifying `Podfile` and our react-native profiles
  - It also seems to be abandoned (hasn't been updated since September 2020)
- [Rugby](https://swiftyfinch.medium.com/rugby-optimise-cocoapods-project-25767b23f287)
  - Builds and caches Pods only
  - Requires an extra build step and bespoke configuration
  - Deviates from how one would consume it normally (in terms of steps and configuration)
- [Module caching](https://medium.com/@SkyscannerEng/hate-to-wait-how-skyscanner-used-module-caching-to-cut-app-build-speed-in-half-e906da1c077e)
  - Cache compiled modules didn't seem to work. Tests crash on startup.
  - Requires building `Pods.xcodeproj` separately from the app

In the end, [Ccache](https://ccache.dev/) is the only one that works consistently, and with minimal effort. It wraps existing compilers (e.g. Clang, GCC), and only requires a configuration. Build steps and so on remain the same. PSPDFKit has written more about their experience here:

  - [Faster Compilation with Ccache 4.0](https://pspdfkit.com/blog/2020/faster-compilation-with-ccache/)
  - [Using Ccache for Fun and Profit](https://pspdfkit.com/blog/2015/ccache-for-fun-and-profit/)

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [ ] Windows

### Test plan

CI should pass and be faster on cache hit.

| iOS cache miss (27m 54s) | iOS cache hit (8m 24s) |
| :-: | :-: |
| image | image |

### DIFF
--- a/.github/actions/setup-toolchain/action.yml
+++ b/.github/actions/setup-toolchain/action.yml
@@ -3,6 +3,8 @@ description: Sets up the toolchain for the project
 inputs:
   platform:
     description: The target platform to set up toolchain for
+  variant:
+    description: Whether we're building from example, template, etc.
 runs:
   using: composite
   steps:
@@ -19,7 +21,7 @@ runs:
       if: ${{ inputs.platform == 'windows' }}
       uses: microsoft/setup-msbuild@v1.1
     - name: Set up Ruby
-      if: ${{ inputs.platform == 'ios' || inputs.platform == 'macos' }}
+      if: ${{ inputs.platform == 'ios' || inputs.platform == 'macos' || inputs.platform == 'node' }}
       uses: ruby/setup-ruby@v1.99.0
       with:
         ruby-version: "3.0"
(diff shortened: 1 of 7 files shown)


tido64:
**iOS**
image

**macOS**
image

.github/actions/setup-toolchain/action.yml
```diff
@@ -28,3 +30,19 @@ runs:
...
+          echo '::set-output name=cache-key::false'
+        fi
+      shell: bash
+    - name: Cache /.ccache
+      if: ${{ steps.cache-key-generator.outputs.cache-key != 'false' }}
+      uses: actions/cache@v2
+      with:
+        path: .ccache
+        key: ${{ runner.os }}-${{ inputs.variant }}-ccache-${{ steps.cache-key-generator.outputs.cache-key }}
+        restore-keys: ${{ runner.os }}-${{ inputs.variant }}-ccache-
```
afoxman, line 48: Is something missing after the last '-'? Or is this supposed to be a key prefix for all keys in that could be generated by this run -- e.g. for reuse, or periodic cleanup?

afoxman (approved):

.github/actions/setup-toolchain/action.yml
tido64: The docs aren't 100% clear on this, but I think your assumption is correct: https://github.com/actions/cache#inputs

PR #761, opened 2022-02-10
ci(nightly): `set-react-version` needs an installed dependency

tido64:
### Description

Since `set-react-version` now depend on `pacote`, we are now required to run `yarn install` before using it.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

n/a

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -130,6 +130,7 @@ jobs:
           # https://github.com/facebook/react-native/issues/30036 and
           # https://github.com/microsoft/react-native-macos/issues/620 for more
           # details.
+          yarn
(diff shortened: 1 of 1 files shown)


kelset (approved):
will this make CI slower?

tido64:
> will this make CI slower?

Somewhat, but not significantly I think. Yarn v3's cache is pretty good and fast.

tido64:
> > will this make CI slower?
> 
> Somewhat, but not significantly I think. Yarn v3's cache is pretty good and fast.

Also, just to be clear: This only affects nightly builds. PRs and mainline builds are unaffected.

PR #762, opened 2022-02-11
feat(windows): add signing capability

tido64:
### Description

Adds signing capability, and refactored project files to support incoming react-native-windows 0.68.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [x] Windows

### Test plan

```
yarn
yarn run set-react-version 0.68
yarn
cd example
yarn install-windows-test-app
start windows/Example.sln
```

Tested that deployment works in `react-native-webview`.


tido64:
If you run `yarn windows`, it will successfully build and launch the app. However, once you open the solution in Visual Studio and try to build, it will consistently fail with: `1>C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Microsoft\NuGet\16.0\Microsoft.NuGet.targets(198,5): error : Your project does not reference "UAP,Version=v10.0" framework. Add a reference to "UAP,Version=v10.0" in the "TargetFrameworks" property of your project file and then re-run NuGet restore.`

Once it's failed, it will be stuck in this state. Running `yarn windows` after this will also fail.

cc @JunielKatarn

example/metro.config.js
```diff
@@ -23,8 +23,14 @@ const blockList = exclusionList([
   // crash if its already running
   new RegExp(`${path.join(__dirname, "windows").replace(/[/\\]+/g, "/")}.*`),
 
+  // Workaround for `EPERM: operation not permitted, lstat '~\midl-MIDLRT-cl.read.1.tlog'`
+  /.*\.tlog/,
+
+  // Prevent Metro from watching temporary files generated by Visual Studio
+  // otherwise it may crash when they are removed when closing a project.
+  /.*\/.vs\/.*/,
```
kelset, line 31: is this something that we should consider upstreaming to the rnx-kit metro config? sounds generic enough that might be good to just have everywhere

windows/ReactTestApp/ReactTestApp.vcxproj
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="16.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="Current" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
kelset, line 2: noob q: are we sure we want to set "current" instead of a fixed version? are we going to incurr in the risk of like, when the newer version will come up that it will get picked and maybe lead to problems? I don't know VS enough to know if this would be an issue

kelset (approved):
overall looks ok, maybe get someone with more windows specific exp to also check it - like @afoxman?

example/metro.config.js
tido64: Yup: https://github.com/microsoft/rnx-kit/pull/1107

windows/ReactTestApp/ReactTestApp.vcxproj
tido64: This is what upstream does. I'm basically just syncing this file with 0.68 right now.

afoxman:
> overall looks ok, maybe get someone with more windows specific exp to also check it - like @afoxman?

It looks good to me, but I too don't know a whole lot about what is going on. I see the code to inject the new MSBuild directives for cert signing, and a bunch of tests and supporting code.

afoxman (approved):

PR #775, opened 2022-02-16
fix(android): set `allprojects.repositories` in project template

tido64:
### Description

Some community modules require `allprojects.repositories` to be set to build correctly.

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

1. Add `@react-native-community/slider` to example app
   ```diff
   diff --git a/example/package.json b/example/package.json
   index 3097cd5..8e50072 100644
   --- a/example/package.json
   +++ b/example/package.json
   @@ -15,6 +15,9 @@
        "start": "react-native start",
        "windows": "react-native run-windows --sln windows/Example.sln"
      },
   +  "dependencies": {
   +    "@react-native-community/slider": "^4.2.0"
   +  },
      "peerDependencies": {
        "react": "~16.8.6 || ~16.9.0 || ~16.11.0 || ~16.13.1 || ~17.0.1 || ~17.0.2",
        "react-native": "^0.0.0-0 || 0.60 - 0.67 || 1000.0.0",
   ```
2. Run `yarn`
3. Build the Android test app
   ```sh
   cd example/android
   ./gradlew clean build
   ```

### DIFF
--- a/example/android/build.gradle
+++ b/example/android/build.gradle
@@ -12,3 +12,14 @@ buildscript {
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
     }
 }
+
+allprojects {
+    repositories {
+        maven {
+            // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
+            url("${rootDir}/../node_modules/react-native/android")
+        }
+        mavenCentral()
+        google()
+    }
+}
--- a/scripts/configure.js
+++ b/scripts/configure.js
@@ -423,7 +423,10 @@ const getConfig = (() => {
               "    repositories {",
               "        maven {",
               "            // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm",
-              `            url("\${rootDir}/${path.join(path.dirname(testAppRelPath), "react-native")}/android")`,
+              `            url("\${rootDir}/${path.posix.join(
+                path.dirname(testAppRelPath),
+                "react-native"
+              )}/android")`,
               "        }",
               "        mavenCentral()",
               "        google()",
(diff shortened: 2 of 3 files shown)


kelset (approved):

example/android/build.gradle
```diff
@@ -12,3 +12,14 @@ buildscript {
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
     }
 }
+
+allprojects {
+    repositories {
+        maven {
+            // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
```
Saadnajmi, line 19: Nit, Since this comment is taken from the upstream issue, is it worth referencing somewhere?

example/android/build.gradle
```diff
@@ -12,3 +12,14 @@ buildscript {
...
+            url("${rootDir}/../node_modules/react-native/android")
```
Saadnajmi, line 20: Nit: Do we have a more generic way to get the node_modules? For FURN I locally have `  url("$rootDir/../../../node_modules/react-native/android")` but I acknowledge our structure is a little messy.

Saadnajmi (approved):

example/android/build.gradle
tido64: That comment actually comes from the Android template: https://github.com/facebook/react-native/blob/f11dcfaea14249b059aea2474ce36a0665140d4f/template/android/build.gradle#L27

example/android/build.gradle
tido64: We have this util function: https://github.com/microsoft/react-native-test-app/blob/e4ea6749532c50fa975a75d08504cc8f81948be5/android/test-app-util.gradle#L48

But if we want to use it here, I think it would be better to make it a value in `dependencies.gradle` to avoid importing more files.

example/android/build.gradle
Saadnajmi: Oh, never mind then!

PR #784, opened 2022-02-24
fix: enable sanitizers in Xcode build settings instead

tido64:
### Description

Enabling sanitizers in scheme causes issues on some setups when running XCTests.

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [ ] Windows

### Test plan

CI should pass.


crazyfraggle (commented):
This change makes sense to me. Only question open is whether we can/should keep the sanitizers to simulator builds only.

tido64:
> This change makes sense to me. Only question open is whether we can/should keep the sanitizers to simulator builds only.

Let's leave it on for all archs for now. It's consistent with what we had previously, and it's a lot easier to loosen than to tighten later.

kelset (approved):

PR #793, opened 2022-03-07
fix: Add clang static analysis flags

Saadnajmi:
### Description

Add a set of flags that performs a hand picked set of clang static analysis compilation flags to the apple platforms. This is enabled for both debug and ship so that the developer testing setup matches the end product as close as possible. This is similar to https://github.com/microsoft/fluentui-react-native/pull/1456 and https://github.com/microsoft/react-native-macos/pull/1014 (Though both only enable the flags on release builds). 

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [ ] Windows

### Test plan

CI should pass.


Saadnajmi:
Tagging @harrieshin and @anandrajeswaran since I can't add reviewers for some reason.

harrieshin (approved):

kelset:
@Saadnajmi looks like the CI is failing on the commit message: https://github.com/conventional-changelog/commitlint/#what-is-commitlint I'm not sure if it's enough to change the title of the PR (since we'll squash&merge) - @tido64 probably knows best.

This should work btw:
```
fix(ios, macos): add clang static analysis flags
```

ios/ReactTestApp/ReactTestApp.common.xcconfig
```diff
@@ -18,3 +18,17 @@ OTHER_LDFLAGS = $(inherited) -fstack-protector-strong
 SDKROOT = iphoneos
 SWIFT_TREAT_WARNINGS_AS_ERRORS = YES
 WARNING_CFLAGS = -Wall
+
+RUN_CLANG_STATIC_ANALYZER = YES
+CLANG_STATIC_ANALYZER_MODE = deep
+
+// Required security settings for production code (do not override at target/project level, with the possible
```
tido64, line 25: Does this mean that it's only required for code that gets shipped to customers? As far as I understand, `ReactTestApp.*.xcconfig` only applies to the `ReactTestApp` target. For dependencies pulled in by CocoaPods, we need to set the flags for each pod [individually](https://guides.cocoapods.org/syntax/podspec.html#pod_target_xcconfig), or inject them into all targets with
[`user_target_xcconfig`](https://guides.cocoapods.org/syntax/podspec.html#user_target_xcconfig). The alternative is to invoke analyze via `xcodebuild` directly.

tido64:
> @Saadnajmi looks like the CI is failing on the commit message: https://github.com/conventional-changelog/commitlint/#what-is-commitlint I'm not sure if it's enough to change the title of the PR (since we'll squash&merge) - @tido64 probably knows best.
> 
> This should work btw:
> 
> ```
> fix(ios, macos): add clang static analysis flags
> ```

That looks about right. I don't mind changing it manually before we merge though.

ios/ReactTestApp/ReactTestApp.common.xcconfig
Saadnajmi: I intended to add the flags here so they get injected into all targets (I.E: The individual native modules that a project's test app is building). It sounds like I actually want to inject them into `user_target_xcconfig`. I'm not sure if `xcodebuild analyze` is doing something different from these flags. @harrieshin / @anandrajeswaran any opinions?

Saadnajmi:
Talked offline, and this doesn't seem to be the approach we want to take, for a few reasons:

1. Adding static analysis to the default setup increases build times drastically. The preference is to have static analysis performed on a separate CI job / pipeline
2. Overriding [user_target_xcconfig](https://guides.cocoapods.org/syntax/podspec.html#user_target_xcconfig) is not recommended by cooapods. It seems we should let each client be in charge of their own static analysis.

There is still a preference for running `xcodebuild analyze` with these explicit xcconfig flags so you know what is tested. so what we _can_ do is change this PR to add a new job / step to add a new "Static Analysis macOS/iOS" check to the CI.

PR #954, opened 2022-06-07
chore: don't publish internal scripts

tido64:
### Description

We should not publish internal scripts.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

CI should pass. Additionally, we should only be publishing the following files:

```
~rnta % npm pack --dry-run --verbose 2>&1 | grep scripts
npm notice 29.1kB  scripts/configure.js
npm notice 1.4kB   scripts/init.js
npm notice 503B    scripts/link.js
npm notice 892B    scripts/patch-cli-platform-android.js
npm notice 1.9kB   scripts/validate-manifest.js
```


package.json
```diff
@@ -35,7 +35,7 @@
     "/ios",
     "/macos",
     "/schema.json",
-    "/scripts",
+    "/scripts/*.js",
```
kelset, line 38: ok so the assumption/rule here is that we are expecting to have all the .js scripts to be "external" and the *.mjs as internal only, right?

I think it's ok, but maybe document it in the contribution docs? I fear we'll forget unless we have some constant reminder somewhere

tido64: That's a good callout. I've updated the contribution docs. Hopefully, our integration tests will also catch missing files.

kelset (approved):

PR #969, opened 2022-06-23
chore: address CVE-2021-42740

tido64:
### Description

For more details, see https://github.com/advisories/GHSA-g4rg-993r-mgx7

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

n/a

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,6 +138,7 @@
   "resolutions": {
     "@commitlint/is-ignored/semver": "^7.3.5",
     "@microsoft/eslint-plugin-sdl/eslint-plugin-react": "^7.26.0",
+    "@react-native-community/cli-tools/shell-quote": "^1.7.3",
     "core-js-compat/semver": "^7.3.5",
     "npm/chalk": "^4.1.0"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -138,6 +138,7 @@
   "resolutions": {
     "@commitlint/is-ignored/semver": "^7.3.5",
     "@microsoft/eslint-plugin-sdl/eslint-plugin-react": "^7.26.0",
+    "@react-native-community/cli-tools/shell-quote": "^1.7.3",
```
kelset, line 141: interesting - do we need to do this because it's not used by CLI v8 or something like that?

kelset (approved):

package.json
tido64: It's fixed upstream, but we have to override it here since it depends on 1.6.1 explicitly. I'm overriding `@react-native-community/cli-tools` explicitly to remember to remove it when we bump to 0.68.

PR #1041, opened 2022-08-05
chore(docs): update the readme

kelset:
### Description

This PR is a follow up from a conversation we had last week about improving the state of the README, and add references to the migrating guide in it.

To make it shorter, I also moved the quick start guide to its [own wiki page](https://github.com/microsoft/react-native-test-app/wiki/Quick-Start). In the migration wiki I've also added a few more notes about certain things we had to do for blur (you can see them at the bottom of [this section](https://github.com/microsoft/react-native-test-app/wiki/Migrate-an-Existing-Test-App#configure-the-test-app)).

### Platforms affected

N/A

### Test plan

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,28 +1,31 @@
 # React Native Test App
 
-[![Open in Visual Studio Code](https://img.shields.io/static/v1?logo=visualstudiocode&label=&message=Open%20in%20Visual%20Studio%20Code&color=007acc&labelColor=444444&logoColor=007acc)](https://vscode.dev/github/microsoft/react-native-test-app)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,29 +1,32 @@
...
 ["Improve all the repos – exploring Microsoft’s DevExp"](https://youtu.be/DAEnPV78rQc?t=499)
-talk by @kelset and @tido64 from React Native Europe 2021.
+talk by [@kelset](https://github.com/kelset) and [@tido64](https://github.com/tido64) from React Native Europe 2021.
 
-For more about
+Read more about
 [the motivation](https://github.com/microsoft/react-native-test-app/wiki#motivation)
 and [the design](https://github.com/microsoft/react-native-test-app/wiki/Design)
-of this tool, you can refer to the wiki.
+of this tool.
```
tido64: Can we rephrase this so that it still makes sense even without the hyperlinks?

tido64: Same for all the new lines below. I like to have proper sentences for screen readers etc.

kelset: ok, I'll take care of it now - thanks, I didn't think about screen readers 👍

kelset: @tido64 check it out now, lmk if it's better

tido64 (approved):
Thanks, this looks a lot better ❤️

PR #997, opened 2022-07-15
fix: declare support for react-native 0.70

tido64:
### Description

Declares support for react-native 0.70.

Blocked by:

- [x] #843
- [x] #1009
- [x] #1044
- [x] #1049

Resolves #996.

### Platforms affected

- [x] Android
- [x] iOS
- [ ] macOS
- [ ] Windows

### Test plan

Use the new script to test all the flags:

```sh
scripts/test-matrix.sh 0.70
```

- [X] Android - Fabric: disabled
- [x] Android - Fabric: **enabled**
- [X] iOS - Fabric: disabled / Hermes: disabled
- [x] iOS - Fabric: disabled / Hermes: **enabled**
- [x] iOS - Fabric: **enabled** / Hermes: disabled
- [x] iOS - Fabric: **enabled** / Hermes: **enabled**
- [ ] ~~macOS~~
- [ ] ~~Windows - NuGet: disabled~~
- [ ] ~~Windows - NuGet: **enabled**~~

### Screenshots

| Platform | App | +Hermes | +Fabric | +Fabric +Hermes |
| --: | :-: | :-: | :-: | :-: |
| Android | n/a | Screenshot_1661187434 | n/a | Screenshot_1661188439 |
| iOS | Simulator Screen Shot - iPhone 13 - 2022-08-22 at 18 57 16 | image | image | image |

PR #1050, opened 2022-08-16
chore: make example app more informational

tido64:
### Description

Display build parameters in the example app for a better overview. This makes it easier to create screenshots for new versions of react-native.

See also #997 for the current issue with screenshots.

The Fabric detector is borrowed from [ScrollViewStickyHeader.js](https://github.com/facebook/react-native/blob/d35aab45b9a2cb79f234fd0d599cc50793866ee2/Libraries/Components/ScrollView/ScrollViewStickyHeader.js#L188-L192)

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

| Android | iOS | iOS with Fabric | macOS |
| :-: | :-: | :-: | :-: |
| Screenshot_1660819993 | image | Simulator Screen Shot - iPhone 13 - 2022-08-16 at 15 29 48 | image |

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -36,6 +36,7 @@
     },
     {
       "matchPackageNames": [
+        "@types/react-native",
         "react-native",
         "react-native-macos",
         "react-native-windows"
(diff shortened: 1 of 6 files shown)


kelset (approved):
LGTM - tested locally and it works like a charm 👍
:shipit: 

Screenshot 2022-08-18 at 12 29 51

PR #1060, opened 2022-08-19
fix(apple): fix blank views after reloading

tido64:
### Description

New React root views are blank after reloading at least once.

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [ ] Windows

### Test plan

1. Build and run the iOS app
2. Shake (^⌘Z) and reload, or toggle "Remote Debugging" (which will trigger a reload)
3. Navigate back
4. Press to open "App"

The new screen should not be blank.

toggle-debugging

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -31,7 +31,7 @@
       "matchPackagePrefixes": ["com.squareup.moshi"]
     },
     {
-      "matchPackageNames": ["react"],
+      "matchPackageNames": ["com.facebook.react:hermes-engine", "react"],
       "enabled": false
     },
     {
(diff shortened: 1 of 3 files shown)


kelset (approved):
interesting - I wonder how you figured this one out 🤣 :shipit:

tido64:
> interesting - I wonder how you figured this one out 🤣 :shipit:

Ah, I kinda remembered that we added a hack to get a fresh bridge instance on reload back in 0.60 or thereabouts when things were more flaky and you could end up with a bridge that was invalidated. Doesn't seem like it's necessary any longer, and it also looks like we get an inner bridge now.

PR #1101, opened 2022-09-11
enhance: add/show default value to init command

### LINKED ISSUE #1100: init command enhancement : add default options to the init command
### Proposal

As a react-native-test-app user while using the init command for the testapp generation i want to get some default values to be considered and shown. 

sometimes bit lazy to type the name or values while doing a POC 😅 . 
`npm init` commands also gives default values which initialisation of project. 

It will look something like this : 
Screenshot 2022-09-11 at 1 27 36 PM

you press enter to take that value or enter the new values to be considered. 

### Alternatives

It doesn't shows the value that exist in the current init command. 

### Implementation Details

```
await require("prompts")([
    {
      type: "text",
      name: "name",
      message: "What is the name of your test app?",
      initial: "testapp",
      validate: Boolean,
    }])
```

**prompts** gives us a 'initial' prop to be pass with is considered as default value and shown as well  like in the above section.

### Additional Context

_No response_

### Code of Conduct

rohanprasadofficial:
### Description

I am enhancing the existing `init` that exist for the initialisation of the RNTA in a react-native repo. 
Added a default value which , user can press enter and continue to select that value.  

Resolves #1100 
### Platforms affected

- [x] Android
- [x] iOS
- [x] macOS
- [x] Windows

### Test plan

1. yarn add react-native-test-app --dev
2. yarn init-test-app

The initialisation prompt will be shown and default value are not shown. 
After this PR the default values will be shown as well as below 

Screenshot 2022-09-11 at 1 27 36 PM

Screenshot 2022-09-11 at 1 27 49 PM

### DIFF
--- a/scripts/init.js
+++ b/scripts/init.js
@@ -18,7 +18,7 @@
       type: "text",
       name: "name",
       message: "What is the name of your test app?",
-      initial: "testapp",
+      initial: "TestApp",
       validate: Boolean,
     },
     {
--- a/scripts/init.js
+++ b/scripts/init.js
@@ -18,6 +18,7 @@
       type: "text",
       name: "name",
       message: "What is the name of your test app?",
+      initial: "testapp",
       validate: Boolean,
     },
     {
@@ -36,6 +37,7 @@
       type: "text",
       name: "packagePath",
       message: "Where should we create the new project?",
+      initial: "example",
       validate: Boolean,
     },
   ]);


scripts/init.js
```diff
@@ -18,6 +18,7 @@
       type: "text",
       name: "name",
       message: "What is the name of your test app?",
+      initial: "testapp",
```
tido64: Since this is the app name, I think we should capitalize it:
```suggestion
      initial: "TestApp",
```

tido64 (approved):
Thanks for adding this 😄

PR #1115, opened 2022-09-13
ci: set up Ccache earlier

tido64:
### Description

Set up Ccache earlier to capture `pod install` artifacts, and allow expansion to Android later.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

CI should pass.

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -1,7 +1,7 @@
 tap 'homebrew/cask-versions'
 brew 'clang-format'
 brew 'ktlint'
-brew 'node'
+brew 'n'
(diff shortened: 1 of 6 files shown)


Brewfile
```diff
@@ -1,7 +1,7 @@
 tap 'homebrew/cask-versions'
 brew 'clang-format'
 brew 'ktlint'
-brew 'node'
+brew 'n'
```
kelset, line 4: ❤️

kelset (approved):

PR #1124, opened 2022-09-19
feat(android): implement QR code scanner

### LINKED ISSUE #1146: Handle random URLs more gracefully
one small thing, that probably needs to be tested separately for both iOS and Android is to handle gracefully when the url is not a metro bundler one. In my first test I generated a random url and the app just crashes. Would be nice to error out, maybe with a "hey this url is not a valid metro instance, try again" or something.

But it's probably something that both iOS and Android need to be coded for, so it can be done as a follow up PR

_Originally posted by @kelset in https://github.com/microsoft/react-native-test-app/pull/1124#pullrequestreview-1130232559_

tido64:
### Description

Implements the QR code scanner feature that has existed on iOS for some time.

Resolves #1146.

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

device-2022-09-19-201856

[CameraX](https://developer.android.com/jetpack/androidx/releases/camera) + [ML Kit](https://developers.google.com/ml-kit/vision/barcode-scanning/android) adds about 4-5 MBs to the installed app size. Make sure that if single app mode is enabled, we don't get that size bump.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -66,6 +66,7 @@ project.ext.react = [
     appName              : getAppName(),
     applicationId        : getApplicationId(),
     architectures        : getArchitectures(),
+    enableCamera         : !getSingleAppMode(),
     enableFabric         : isFabricEnabled(project),
     enableFlipper        : getFlipperVersion(rootDir),
     enableHermes         : true,
@@ -258,6 +259,12 @@ android {
     }
 
     sourceSets {
+        if (project.ext.react.enableCamera) {
+            main.java.srcDirs += "src/camera/java"
+        } else {
+            main.java.srcDirs += "src/no-camera/java"
+        }
+
         if (project.ext.react.enableFlipper) {
             debug.java.srcDirs += "src/flipper/java"
         }
@@ -349,6 +356,12 @@ dependencies {
     testImplementation libraries.mockitoInline
     androidTestImplementation libraries.androidJUnit
     androidTestImplementation libraries.androidEspressoCore
+
+    if (project.ext.react.enableCamera) {
+        implementation libraries.androidCamera
+        implementation libraries.androidCameraMlKitVision
+        implementation libraries.mlKitBarcodeScanning
+    }
 
     if (project.ext.react.enableFlipper) {
         def flipperVersion = project.ext.react.enableFlipper
--- a/android/app/src/camera/java/com/microsoft/reacttestapp/camera/QRCodeScannerFragment.kt
+++ b/android/app/src/camera/java/com/microsoft/reacttestapp/camera/QRCodeScannerFragment.kt
@@ -1,0 +1,99 @@
+package com.microsoft.reacttestapp.camera
+
+import android.os.Bundle
+import android.os.Looper
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.camera.core.ImageCapture.CAPTURE_MODE_MINIMIZE_LATENCY
+import androidx.camera.mlkit.vision.MlKitAnalyzer
+import androidx.camera.view.CameraController.COORDINATE_SYSTEM_VIEW_REFERENCED
+import androidx.camera.view.LifecycleCameraController
+import androidx.camera.view.PreviewView
+import androidx.core.os.HandlerCompat
+import androidx.core.util.Consumer
+import androidx.fragment.app.DialogFragment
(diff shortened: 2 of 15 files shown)


android/app/src/camera/java/com/microsoft/reacttestapp/camera/QRCodeScannerFragment.kt
```diff
@@ -0,0 +1,99 @@
...
+
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View {
+        val view = inflater.inflate(R.layout.camera_view, container, false)
+        previewView = view.findViewById(R.id.preview_view)
+
+        val barcodeScanner = BarcodeScanning.getClient(
```
arazabishov: Nit: you can override `onCreate()` and move BarcodeScanner construction there. `onCreateView` can be called more frequently than `onCreate`.

android/app/src/camera/java/com/microsoft/reacttestapp/camera/QRCodeScannerFragment.kt
```diff
@@ -0,0 +1,99 @@
...
+            MlKitAnalyzer(
+                listOf(barcodeScanner),
+                COORDINATE_SYSTEM_VIEW_REFERENCED,
+                cameraExecutor
+            ) {
+                it.getValue(barcodeScanner)?.let { barcodes ->
+                    for (barcode in barcodes) {
+                        when (barcode.valueType) {
+                            Barcode.TYPE_URL -> {
+                                barcodeScanner.close()
```
arazabishov: Is barcode scanner required to be closed right after being used? I am thinking we could free it up in `onDestroy` instead.

android/app/src/camera/java/com/microsoft/reacttestapp/camera/QRCodeScannerFragment.kt
```diff
@@ -0,0 +1,99 @@
...
+                                val handler = HandlerCompat.createAsync(Looper.getMainLooper())
```
arazabishov: Could you please lift handler creation either to onCreate or declare and assign it to the field?

android/app/src/camera/java/com/microsoft/reacttestapp/camera/QRCodeScannerFragment.kt
```diff
@@ -0,0 +1,99 @@
...
+                                handler.post {
+                                    this.dismiss()
+                                    consumer.accept(barcode.url?.url)
+                                }
+                                break
```
arazabishov: I think break is not needed here :P

arazabishov (approved):

android/app/src/camera/java/com/microsoft/reacttestapp/camera/QRCodeScannerFragment.kt
tido64: It needs to be closed otherwise we'll keep getting results. Added a comment to clarify.

android/app/build.gradle
```diff
@@ -256,6 +257,12 @@ android {
     }
 
     sourceSets {
+        if (project.ext.react.enableCamera) {
+            main.java.srcDirs += "src/camera/java"
+        } else {
+            main.java.srcDirs += "src/no-camera/java"
+        }
```
kelset, line 266: this is pretty cool tbh :D

kelset (approved):
managed to test it via the virtual camera in Android emulator, and overall it works well 👍

one small thing, that probably needs to be tested separately for both iOS and Android is to handle gracefully when the url is not a metro bundler one. In my first test I generated a random url and the app just crashes. Would be nice to error out, maybe with a "hey this url is not a valid metro instance, try again" or something.

But it's probably something that both iOS and Android need to be coded for, so it can be done as a follow up PR

tido64:
> managed to test it via the virtual camera in Android emulator, and overall it works well 👍
> 
> one small thing, that probably needs to be tested separately for both iOS and Android is to handle gracefully when the url is not a metro bundler one. In my first test I generated a random url and the app just crashes. Would be nice to error out, maybe with a "hey this url is not a valid metro instance, try again" or something.
> 
> But it's probably something that both iOS and Android need to be coded for, so it can be done as a follow up PR

https://github.com/microsoft/react-native-test-app/issues/1146

PR #1127, opened 2022-09-21
feat(android): populate home screen with AppRegistry entries

### LINKED ISSUE #189: Populate components list from AppRegistry if app.json is empty/missing

tido64:
### Description

Similar to #482, but for Android.

Resolves #189.

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

See test plan in #482.

~~Flipper crashes on startup with this change. I'm not sure why, but downgrading it to 0.105 resolves the issue.~~ The `SIGILL` comes from the device trying to load different `.so` files on startup, but the app should load fine. Bump to 0.164 to make it go away.

This needs to be tested on:

- [x] 0.64 (no `libjsi.so`, should show blank screen)
- [x] 0.68 (has release `libjsi.so`)
- [x] 0.69 (has debug/release `libjsi.so`)
- [x] 0.69 + new arch (uses `ndkBuild`, `libjsi.so` already linked)
- [x] 0.70 (has debug/release `libjsi.so`)
- [x] 0.70 + new arch (uses `cmake`, `libjsi.so` already linked)

Note that for new arch, you will need to force resolve rn-cli to 9.x for autolinking.

device-2022-09-21-145225

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -158,6 +158,15 @@ android {
                     abiFilters(*project.ext.react.architectures)
                 }
             }
+        } else {
+            externalNativeBuild {
+                cmake {
+                    arguments "-DANDROID_STL=c++_shared",
+                              "-DREACT_COMMON_DIR=${reactNativePath}/ReactCommon",
+                              "-DREACT_JNILIBS_DIR=${buildDir}/outputs/jniLibs"
+                    cppFlags "-std=c++17"
+                }
+            }
         }
     }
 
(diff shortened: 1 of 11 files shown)


kelset (commented):
Tested:

- ❌ 0.64 (no libjsi.so)
  * fails on `yarn android` with error `react-native-test-app/example/node_modules/react-native-test-app/android/app/src/main/java/com/microsoft/reacttestapp/react/TestAppReactNativeHost.kt: (181, 24): Unresolved reference: loadLibrary` and
	   ```
		   Execution failed for task ':app:compileDebugKotlin'.
		> A failure occurred while executing org.jetbrains.kotlin.compilerRunner.GradleCompilerRunnerWithWorkers$GradleKotlinCompilerWorkAction
		   > Compilation error. See log for more details
	   ```
   
- ✅ 0.68 (has release libjsi.so)
- ✅🟡 0.69 (has debug/release libjsi.so) (but for some reason it just shows a blank page instead of the normal one with all the info, and just the version in the bottom right corner)
- ❌ 0.69 + new arch (uses ndkBuild, libjsi.so already linked)
  * fails on `yarn android` with error `react-native-test-app/example/android/app/build/generated/rncli/src/main/jni/Android-rncli.mk: No such file or directory` and `make: *** No rule to make target '/Users/lsciandra/Developer/oss/react-native-test-app/example/android/app/build/generated/rncli/src/main/jni/Android-rncli.mk'.  Stop.`
- ✅ 0.70 (has debug/release libjsi.so)
- ✅ 0.70 + new arch (uses cmake, libjsi.so already linked)

tido64:
@kelset:

For 0.64, we need a fix that landed on `trunk`. I've rebased now so it should work.

For 0.69: I forgot to mention that you need RN-CLI 9.x because 8 does not yet have autolinking. I also had to remove `react-native-safe-area-context` because the latest version dropped support for 0.69 and below when on new arch. Otherwise, I cannot repro the white screen you're seeing.

| | |
|-|-|
| Screenshot_1664903958 | Screenshot_1664906394 |

kelset:
yeah, with the adjustments I managed to make it work for both 0.64 and both versions of 0.69 🎉

LGTM, when you want to open for review I can ✅

kelset (approved):
:shipit:

PR #1145, opened 2022-10-04
fix(ios): crash on startup on device because sanitizers are enabled

tido64:
### Description

To enable sanitizers on device, it's not enough to enable them in config. We need to enable them in the build scheme instead:

    Edit Scheme… > Run > Diagnostics > Runtime Sanitization

image

### Platforms affected

- [ ] Android
- [x] iOS
- [ ] macOS
- [ ] Windows

### Test plan

```
cd example
open ios/Example.xcworkspace
```

Plug in an iOS device and target it. The app should not crash on startup.

Target a simulator and make sure that it also doesn't crash on startup.

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -285,7 +285,7 @@ PODS:
   - React-jsinspector (0.68.3)
   - React-logger (0.68.3):
     - glog
-  - react-native-safe-area-context (4.3.4):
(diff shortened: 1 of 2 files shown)


kelset (commented):
can confirm with this fix, the example app runs correctly on a real device 👍

kelset (approved):

PR #1162, opened 2022-10-18
chore: prevent Renovate from bumping `androidx.appcompat`

tido64:
### Description

For react-native <0.68, `androidx.appcompat:appcompat` is pinned to 1.3.1 to prevent crashes. Renovate tries to bump it here: https://github.com/microsoft/react-native-test-app/pull/1150/files#diff-9526ccfd1d1813ed49c39f8c54dbeb512607376a007d824b905bc8b4e4d202d9

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

```
npm run set-react-version 0.67
yarn
cd example
yarn android
```

Screenshot_1666086573

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -374,7 +374,7 @@ dependencies {
         // androidx.appcompat:appcompat:1.4.0+ breaks TextInput. This was fixed
         // in react-native 0.68. For more details, see
         // https://github.com/facebook/react-native/issues/31572.
-        implementation "androidx.appcompat:appcompat:1.3.1"
+        implementation(["androidx.appcompat", "appcompat", "1.3.1"].join(":"))
     } else {
         implementation libraries.androidAppCompat
     }


android/app/build.gradle
```diff
@@ -374,7 +374,7 @@ dependencies {
         // androidx.appcompat:appcompat:1.4.0+ breaks TextInput. This was fixed
         // in react-native 0.68. For more details, see
         // https://github.com/facebook/react-native/issues/31572.
-        implementation "androidx.appcompat:appcompat:1.3.1"
+        implementation(["androidx.appcompat", "appcompat", "1.3.1"].join(":"))
```
kelset, line 377: isn't there a way to touch some Renovate config instead of this workaround?

tido64: We still want it to upgrade the other instance (in [`dependencies.gradle`](https://github.com/microsoft/react-native-test-app/blob/bf952e4e4ae0fe9662beb2a3dd88cfd064d9a560/android/dependencies.gradle#L36)). Not aware of any configuration that would allow pinning one instance and not the other.

kelset: ok

kelset (approved):

PR #1164, opened 2022-10-21
ci: `set-output` is deprecated

tido64:
### Description

More information here: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Also took the opportunity to update the `rnx-build.yml` workflow from rnx-kit.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

CI should pass.

### DIFF
--- a/.github/workflows/rnx-build.yml
+++ b/.github/workflows/rnx-build.yml
@@ -226,7 +226,7 @@ jobs:
           appx_manifest=$(find ${{ github.event.inputs.architecture }}/Debug -name AppxManifest.xml -type f | head -1)
           app_name=$(basename $(dirname ${appx_manifest}))
           cp ${appx_manifest} AppPackages/${app_name}/*
-          echo "::set-output name=app-name::${app_name}"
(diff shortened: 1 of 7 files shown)


.github/workflows/rnx-build.yml
```diff
@@ -10,6 +10,10 @@ on:
         description: "Supported device types are `device`, `emulator`, `simulator`"
         required: true
         default: "simulator"
+      distribution:
```
afoxman, line 13: Looks like a few changes in this file snuck into the PR. I see 2 instances of ::set-output but the rest should probably go in a different PR.

tido64: This is just a copy from rnx-kit. I can separate it out to a different PR though.

afoxman: Makes sense. If its intentional, than maybe update the PR description or something like that.

afoxman (approved):

PR #1167, opened 2022-10-25
feat: make `init` command standalone

tido64:
### Description

Make `init` command standalone. The command should be able to create a new app using the already installed copy of `react-native`, or download one if necessary.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

- Verify that it uses the installed `react-native`: `node scripts/init.js`
- Verify that it downloads `react-native@0.69`: `node scripts/init.js --version 0.69`
- Verify that it determines the latest version and downloads it:
  ```
  # start outside any repository
  node /path/to/react-native-test-app/scripts/init.js
  ```

For each of the test, verify that the generated `package.json` contains the correct versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
   "main": "scripts/configure.js",
   "bin": {
     "configure-test-app": "scripts/configure.js",
+    "init": "scripts/init.js",
     "init-test-app": "scripts/init.js",
     "install-windows-test-app": "windows/test-app.js"
   },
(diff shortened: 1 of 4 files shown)


kelset (approved):
Looks great, tested all cases locally and I think we can merge it as is :shipit: 

A couple of things I noticed and that we can follow up on as separate PRs
- it puts as RNTA version in the package.json the one from branch, so `"^0.0.1-dev"` this is probably because I'm running it locally, let's verify post release that it sets to latest available on npm
- doesn't do pod install, so might be good to add as a step when it detects that the project targets iOS
- exposing Hermes flag: neither iOS nor Android side I could see an easy way to switch on/off Hermes. Maybe it's just me not looking in the right place

PR #1168, opened 2022-10-26
fix(android): bump NDK to r23c

tido64:
### Description

Bumps Android NDK to r23c when using AGP 7.2.2 or older.

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

Make sure the Android app builds with 0.64 and latest 0.70.

| 0.64 | 0.70 |
| :-: | :-: |
| Screenshot_1666775910 | Screenshot_1666776310 |

### DIFF
--- a/android/dependencies.gradle
+++ b/android/dependencies.gradle
@@ -10,11 +10,12 @@ ext {
     targetSdkVersion = 29
 
     def reactNativeVersion = getPackageVersionNumber("react-native", rootDir)
+    def useAndroidPlugin_7_2 = reactNativeVersion > 0 && reactNativeVersion < 7100
 
     // We need only set `ndkVersion` when building react-native from source.
     if (rootProject.hasProperty("ANDROID_NDK_VERSION")) {
         ndkVersion = rootProject.properties["ANDROID_NDK_VERSION"]
-    } else if (System.properties["os.arch"] == "aarch64" && reactNativeVersion < 7100) {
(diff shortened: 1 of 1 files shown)


android/dependencies.gradle
```diff
@@ -9,16 +9,18 @@ ext {
...
     if (rootProject.hasProperty("ANDROID_NDK_VERSION")) {
         ndkVersion = rootProject.properties["ANDROID_NDK_VERSION"]
-    } else if (System.properties["os.arch"] == "aarch64") {
-        // Android NDK added support for Apple M1 in r24:
-        // https://github.com/android/ndk/wiki/Changelog-r24
-        ndkVersion = "24.0.8215888"
+    } else if (System.properties["os.arch"] == "aarch64" || reactNativeVersion >= 7100) {
+        // NDK r23c has been patched to support Apple M1 and will be default in
+        // 0.71: https://github.com/facebook/react-native/pull/35066
+        ndkVersion = "23.2.8568313"
```
tido64: This is the latest r23c. Not quite sure why core is using a slightly older version: https://github.com/facebook/react-native/pull/35066/files#r1005429823

tido64: We probably want to use the version that is default in AGP. As of 7.3, the NDK version is 23.1.7779620, which should include M1 patches: https://developer.android.com/studio/releases/gradle-plugin#compatibility-7-3-0

We should not set this prop if we are on 7.3 or later.

kelset (approved):

PR #1169, opened 2022-11-02
fix(android): fix compatibility with react-native@nightly

tido64:
### Description

Fixes for building with Android nightlies.

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

Repeat the steps below for 0.64 and 0.70:

```
npm run set-react-version <version>
npm run clean && yarn
cd example
yarn android
```

To build nightly, specify `nightly` for version. Note however, that you will also need to remove `react-native-safe-area-context` as it does not build.

| 0.64 | 0.70 | nightly |
| :-: | :-: | :-: |
| Screenshot_1667415909 | Screenshot_1667415756 | Screenshot_1667419700 |

### DIFF
--- a/android/app/src/main/java/com/microsoft/reacttestapp/react/TestAppReactNativeHost.kt
+++ b/android/app/src/main/java/com/microsoft/reacttestapp/react/TestAppReactNativeHost.kt
@@ -162,7 +162,6 @@ class TestAppReactNativeHost(
             .addPackages(packages)
             .setUseDeveloperSupport(useDeveloperSupport)
             .setInitialLifecycleState(LifecycleState.BEFORE_CREATE)
-            .setUIImplementationProvider(uiImplementationProvider)
             .setRedBoxHandler(redBoxHandler)
             .setJSIModulesPackage(jsiModulePackage)
             .build()
--- a/android/test-app-util.gradle
+++ b/android/test-app-util.gradle
@@ -223,14 +223,24 @@ ext.getVersionName = {
 }
 
 ext.isFabricEnabled = { project ->
-    return isNewArchitectureEnabled(project) ||
-        (isPropertySet(project, "USE_FABRIC", "1") &&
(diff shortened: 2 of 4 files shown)


android/app/src/main/java/com/microsoft/reacttestapp/react/TestAppReactNativeHost.kt
```diff
@@ -162,7 +162,6 @@ class TestAppReactNativeHost(
             .addPackages(packages)
             .setUseDeveloperSupport(useDeveloperSupport)
             .setInitialLifecycleState(LifecycleState.BEFORE_CREATE)
-            .setUIImplementationProvider(uiImplementationProvider)
```
tido64, line 165: FYI, this was introduced in https://github.com/facebook/react-native/commit/b002df945b267c6e36aa1041397ffd1c73167037, and subsequently removed in https://github.com/facebook/react-native/commit/e7d75631959f1ce3f50a44f13056eacaf82378ff. We've always set the default value here so removing it doesn't affect anything.

android/test-app-util.gradle
```diff
@@ -223,14 +223,24 @@ ext.getVersionName = {
 }
 
 ext.isFabricEnabled = { project ->
-    return isNewArchitectureEnabled(project) ||
-        (isPropertySet(project, "USE_FABRIC", "1") &&
-            getPackageVersionNumber("react-native", project.rootDir) >= 6800)
+    if (isNewArchitectureEnabled(project)) {
```
kelset, line 226: btw I have mentioned it somewhere, but from 0.71 there are going to be new "togglers" for Fabric and TM: 
* https://github.com/facebook/react-native/pull/35091
* https://github.com/facebook/react-native/pull/35117
don't know if you want to add this separately when trying out 0.71 rc0

kelset (approved):

PR #1197, opened 2022-11-10
fix(ios): add support for `react-native-reanimated`

### LINKED ISSUE #1195: iOS: Unable to install reanimated
### Description
Adding `"react-native-reanimated": "2.10.0"` to the example app, installing pods and running the app leads to 
```
TypeError: undefined is not an object (evaluating 'this.InnerNativeModule.installCoreFunctions')
```
even after adding the reanimated [babel plugin](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/installation#babel-plugin) and running metro with cache cleared. Looks like the JSI is not linked/installed.

Is this something expected atm? Didn't see it reported among known issues [https://github.com/microsoft/react-native-test-app/wiki/Troubleshooting](https://github.com/microsoft/react-native-test-app/wiki/Troubleshooting)

tido64:
### Description

Added support for injecting code into the React bridge delegate, allowing somewhat safer installation of the Reanimated JSI executor.

Resolves #1195.

### Platforms affected

- [ ] Android
- [x] iOS
- [ ] macOS
- [ ] Windows

### Test plan

```diff
diff --git a/example/App.js b/example/App.js
index c9cd14d..b3bce70 100644
--- a/example/App.js
+++ b/example/App.js
@@ -10,6 +10,7 @@ import {
   useColorScheme,
   View,
 } from "react-native";
+import Animated from "react-native-reanimated";
 import { SafeAreaProvider, SafeAreaView } from "react-native-safe-area-context";
 // @ts-expect-error
 import { version as coreVersion } from "react-native/Libraries/Core/ReactNativeVersion";
@@ -179,6 +180,16 @@ const App = ({ concurrentRoot }) => {
         >
           <Header />
           <DevMenu />
+          <Animated.View
+            style={[
+              {
+                width: 100,
+                height: 80,
+                backgroundColor: "orange",
+                margin: 30,
+              },
+            ]}
+          />
           <View style={styles.group}>
             <Feature value={getReactNativeVersion()}>React Native</Feature>
             <Separator />
diff --git a/example/app.json b/example/app.json
index ef69fa7..c492b30 100644
--- a/example/app.json
+++ b/example/app.json
@@ -13,6 +13,9 @@
       "presentationStyle": "modal"
     }
   ],
+  "plugins": [
+    "react-native-test-app/plugins/reanimated.js"
+  ],
   "resources": {
     "android": [
       "dist/res",
diff --git a/example/babel.config.js b/example/babel.config.js
index 5ae49c3..8a4543d 100644
--- a/example/babel.config.js
+++ b/example/babel.config.js
@@ -1,3 +1,4 @@
 module.exports = {
   presets: ["module:metro-react-native-babel-preset"],
+  plugins: ["react-native-reanimated/plugin"],
 };
diff --git a/example/package.json b/example/package.json
index a51d4cf..e2230e2 100644
--- a/example/package.json
+++ b/example/package.json
@@ -24,6 +24,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.0.0",
+    "@expo/config-plugins": "^5.0.0",
     "@types/jest": "^27.0.0",
     "@types/react-native": "^0.68.0",
     "jest": "^27.0.0",
@@ -32,6 +33,7 @@
     "react": "17.0.2",
     "react-native": "^0.68.2",
     "react-native-macos": "^0.68.3",
+    "react-native-reanimated": "^2.0",
     "react-native-safe-area-context": "^4.3.4",
     "react-native-test-app": "workspace:.",
     "react-native-windows": "^0.68.8"
```


plugins/reanimated.js
```diff
@@ -0,0 +1,73 @@
+const { createRunOncePlugin } = require("@expo/config-plugins");
+const {
+  mergeContents,
+} = require("@expo/config-plugins/build/utils/generateCode");
+const { withBridgeDelegate } = require("./index");
+
+const NAME = "react-native-reanimated";
+
+function addContents(tag, src, newSrc, anchor) {
```
rrebase, line 22: The plugin isn't written in TS due to the overhead of dealing with the transpiling step here?

rrebase (commented):
Tested it out with a different setup and works well 🙌

plugins/reanimated.js
tido64: In this repo, we use [TSDoc](https://tsdoc.org/) + [`@ts-check`](https://www.typescriptlang.org/docs/handbook/intro-to-js-ts.html), and run `tsc` on CI. So we do type checking, but not transpilation. Mostly to make it easier to debug and hack on the consumer side. I will add types and do some more testing before making this ready for review.

kelset (approved):
On one hand, I'm as always impressed by the tricks you can pull out of the hat. Very interesting approach to addressing this - I've tested also locally and it works as expected:
Screenshot 2022-11-29 at 14 54 33

My current concern is that this type of workaround is very much custom per each plugin - do we have any docs on this process? Can we make sure that if this approach is the one we want folks to use (and I mean even other devs implementing support for other plugins) is documented at least partially?

tido64:
I will follow up with some docs. I was supposed to document the config-plugins feature, but never got around to it.

PR #1231, opened 2022-11-28
fix: workaround for npm ignoring `.gitignore`

### LINKED ISSUE #1228: bug: ENOENT: no such file or directory, copyfile 'node_modules/react-native-test-app/example/.gitignore' -> 'example/.gitignore'
### What happened?

running `init-test-app` crashes due to behavior in npm. i'm no [expert on the matter](https://github.com/npm/npm/issues/1862) so take this solution with grain of salt. adding a `.npmignore` file will preserve the `.gitignore` file in the example folder. steps to reproduce are simple enough. i'm using node 18.9.1 with npm v8.19.1. run `npm install react-native-test-app` and you will see that under node_modules/react-native-test-app/example there is a .npmignore file but not the .gitignore file as expected. if you download the tarball from npm you will see the .gitignore as expected. simplest solution is to just add an empty .npmignore file to the example folder to prevent this behavior 🤷‍♂️ 

### Version

2.0.3

### What platforms are you seeing this issue on?

- [ ] Android
- [ ] iOS
- [X] macOS
- [ ] Windows

### System Information

```true
not relevant
```

### Steps to Reproduce

listed in description above ↑ 

### Code of Conduct

tido64:
### Description

`npm install` seems to be renaming `.gitignore` files to `.npmignore`. This breaks our templates. As a workaround, we'll rename the files ourselves, from `.gitignore` to `_gitignore`, to avoid any further attempts to mutate the package.

Resolves #1228.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

```
npm pack
tar tf react-native-test-app-0.0.1-dev.tgz | grep gitignore
```

Verify that `.gitignore` files have been renamed to `_gitignore`, i.e.:

```
% tar tf react-native-test-app-0.0.1-dev.tgz | grep ignore
package/example/_gitignore
package/example/windows/_gitignore
```


kelset (approved):
tested locally, works as expected 👍 kind of surprised by this behaviour but I guess that's why in [RN's template](https://github.com/facebook/react-native/tree/main/template) we have all the dotenv files as "_"

PR #1255, opened 2023-01-05
chore: `set-react-version` should remove packages

tido64:
### Description

`set-react-version` should remove packages if the specified version does not exist.

Trying to target, say 0.71, can fail because react-native-macos 0.68 is still installed.

It's also currently the cause of failing nightlies: https://github.com/microsoft/react-native-test-app/actions/runs/3843735554

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

`react-native-macos` 0.71 currently does not exist. Run `npm run set-react-version 0.71` and verify that `react-native-macos` was removed from both `package.json` and `example/package.json`.

### DIFF
--- a/scripts/set-react-version.mjs
+++ b/scripts/set-react-version.mjs
@@ -251,14 +251,12 @@ if (!isValidVersion(version)) {
 
   const manifests = ["package.json", "example/package.json"];
   for (const manifestPath of manifests) {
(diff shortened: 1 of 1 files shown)


kelset (approved):
LGTM, tested & ✅
:shipit:

PR #1260, opened 2023-01-10
fix(apple): Don't ignore warnings in 'React-Core'

Saadnajmi:
### Description

#1259 added code to ignore some clang warnings. https://github.com/microsoft/react-native-macos/pull/1612 should fix those warnings so we no longer need to ignore them.

### Platforms affected

- [ ] Android
- [ ] iOS
- [x] macOS
- [ ] Windows

### Test plan

```
npm run set-react-version canary-macos
yarn
cd example
../scripts/xcodebuild.sh macos/Example.xcworkspace build-for-testing
```

### DIFF
--- a/ios/ReactTestApp/React+Fabric.mm
+++ b/ios/ReactTestApp/React+Fabric.mm
@@ -1,8 +1,5 @@
 #import "React+Fabric.h"
 
-#pragma clang diagnostic push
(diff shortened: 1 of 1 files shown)


tido64 (approved):

tido64:
Thanks for fixing those warnings!

PR #1293, opened 2023-02-08
fix: further simplify project config

tido64:
### Description

Further simplify `react-native.config.js`

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

- 0.68: The output of `node -e "console.dir(require('./react-native.config.js'), {depth: null})"` should not differ before and after this change.
- 0.69+: `yarn react-native config` should not fail

Also tested in AsyncStorage with:

```js
    const { configureProjects } = require("react-native-test-app");
    return configureProjects({
      android: {
        sourceDir: path.join('example', 'android'),
      },
      ios: {
        sourceDir: path.join('example', 'ios'),
      },
      windows: {
        sourceDir: path.join('example', 'windows'),
        solutionFile: path.join('example', 'windows', 'AsyncStorageExample.sln'),
      },
    });
```

### DIFF
--- a/example/react-native.config.js
+++ b/example/react-native.config.js
@@ -1,25 +1,18 @@
 const project = (() => {
-  const fs = require("fs");
-  const path = require("path");
   try {
-    const {
-      androidManifestPath,
-      iosProjectPath,
-      windowsProjectPath,
-    } = require("react-native-test-app");
-    const iosProject = iosProjectPath("ios");
-    return {
+    const { configureProjects } = require("react-native-test-app");
+    return configureProjects({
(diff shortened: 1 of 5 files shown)


tido64:
@kelset: I think I ironed out all the issues. Would appreciate a review 😘

kelset (approved):
sorry it took me so long to get back to this 🙏

tido64:
> sorry it took me so long to get back to this 🙏

No worries! Thanks for reviewing, as always :heart:

PR #1308, opened 2023-02-13
fix(ios): support multiwindow on iPad

lyzhan7:
### Description

Update the [UIApplicationSupportsMultipleScenes](https://developer.apple.com/documentation/bundleresources/information_property_list/uiapplicationscenemanifest/uiapplicationsupportsmultiplescenes?language=objc) from false to true in the plist so that the test app can support multiple windows on iPad.

Consideration of multiwindow scenarios came up in this [PR](https://github.com/microsoft/fluentui-react-native/pull/2614) in FURN. I was able turn on the multiwindow setting locally for the FURN test app, but thought we could turn it on for the test app in general.

### Platforms affected

- [ ] Android
- [x] iOS
- [ ] macOS
- [ ] Windows

### Test plan

Tested by making this change locally in the FURN test app (fluentui-react-native/node_modules/react-native-test-app/ios/ReactTestApp/Info.plist)

Before:

https://user-images.githubusercontent.com/78454019/218600063-afe3bc24-5260-4455-89ab-9dc653708a71.mov

After:

https://user-images.githubusercontent.com/78454019/218600061-23e27b22-00d4-458b-a8a9-0327942bcd2c.mov

Note: I also tried to test directly in the react-native-test-app repo - I thought that after making this change, when I followed the [steps](https://github.com/microsoft/react-native-test-app/wiki/Quick-Start) to generate a test app, that test app would have UIApplicationSupportsMultipleScenes set to true, but every time the setting is still false. If there is something I'm missing please let me know.

### DIFF
--- a/ios/ReactTestApp/Info.plist
+++ b/ios/ReactTestApp/Info.plist
@@ -57,7 +57,7 @@
 	<key>UIApplicationSceneManifest</key>
 	<dict>
 		<key>UIApplicationSupportsMultipleScenes</key>
-		<false/>
+		<true/>
 		<key>UISceneConfigurations</key>
 		<dict>
 			<key>UIWindowSceneSessionRoleApplication</key>


Saadnajmi (approved):
(Not that my review means anything here)

tido64 (approved):
Nice! Thank you :heart:

PR #1363, opened 2023-03-27
feat: use `@rnx-kit/react-native-host`

tido64:
### Description

React Native hosting bits have been extracted to a separate package.

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [ ] Windows

### Test plan

CI should build. Test apps on affected platforms should successfully boot.

#### iOS

```
cd example
yarn ios
```

#### macOS

```
cd example
yarn macos
```

#### iOS + New Architecture

```
npm run set-react-version 0.72
cd example
RCT_NEW_ARCH_ENABLED=1 pod install --project-directory=ios
yarn ios
```


ios/ReactTestApp/ContentView.swift
```diff
@@ -53,7 +53,7 @@ final class ContentViewController: UITableViewController {
 
     override public func motionEnded(_: UIEvent.EventSubtype, with event: UIEvent?) {
         guard event?.subtype == .motionShake,
-              let bridge = reactInstance.bridge,
+              let bridge = reactInstance.host?.bridge,
```
Saadnajmi, line 56: Any reason not to follow the same structure as lines 196-197?
```
guard let host = reactInstance.host,
      let bridge = host.bridge,
```

Saadnajmi (approved):

ios/ReactTestApp/ContentView.swift
tido64: I normally don't create a separate variable if it's not used outside of the block.

kelset (approved):
all works 👍

a couple of small hiccups but nothing strictly related to this PR, so we can proceed with it

tido64:
> all works 👍
> 
> a couple of small hiccups but nothing strictly related to this PR, so we can proceed with it

Thanks ❤️

PR #1367, opened 2023-04-02
fix: replace yargs with minimist to reduce footprint

tido64:
### Description

We don't use everything that yargs has to offer so it brings a bit of a overhead for us. We can replace the bits we need with `minimist` and using `cliui` directly. In the future, we will be able to use [`util.parseArgs`](https://nodejs.org/api/util.html#utilparseargsconfig) and no longer depend on `minimist`.

[Composition](https://bundlephobia.com/package/react-native-test-app@2.3.14)

### Platforms affected

- [x] Android
- [x] iOS
- [x] macOS
- [x] Windows

### Test plan

`scripts/configure.js` and `windows/test-app.js` are the two scripts affected by this change.

- Verify that `--help` is correctly formatted in both big and small terminals
- From the `example` folder, you can run either command to verify that they are getting the arguments passed, for example:
  - `node ../scripts/configure.js` alone should list files that affect all platforms
  - `node ../scripts/configure.js -p android -p ios` should only list Android and iOS files
  - `node ../scripts/configure.js -p windows` should only list Windows files
  - `node ../windows/test-app.js` should generate Visual Studio solution files under `example/windows` (even on a Mac)
  - `node ../windows/test-app.js --autolink` on a Mac will fail because autolinking requires Windows

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,11 +80,12 @@
   "dependencies": {
     "ajv": "^8.0.0",
     "chalk": "^4.1.0",
+    "cliui": "^8.0.0",
     "fast-xml-parser": "^4.0.0",
+    "minimist": "^1.2.8",
     "prompts": "^2.4.0",
     "semver": "^7.3.5",
-    "uuid": "^8.3.2",
-    "yargs": "^16.0.0"
+    "uuid": "^8.3.2"
   },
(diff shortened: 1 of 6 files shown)


kelset (approved):
:shipit:

PR #1390, opened 2023-04-13
Fix compiling SceneDelegate with Xcode 14.3

fortmarek:
### Description

When compiling the `react-native-test-app` with `react-native` 0.71.6 and Xcode 14.3 for iOS, I started to get the following error:
```
❌  /some-path/fixture/node_modules/.generated/ios/ReactTestApp/SceneDelegate.swift:54:21: heterogeneous collection literal could only be inferred to '[String : Any]'; add explicit type annotation if this is intentional

            object: [
                    ^
```

The fix in this PR just adds `as [String: Any]` as the error message suggests.

### Platforms affected

- [ ] Android
- [x] iOS
- [ ] macOS
- [ ] Windows

### Test plan

Try `react-native-test-app` with `react-native` 0.71.6 and Xcode 14.3

### DIFF
--- a/ios/ReactTestApp/SceneDelegate.swift
+++ b/ios/ReactTestApp/SceneDelegate.swift
@@ -54,7 +54,7 @@ final class SceneDelegate: UIResponder, UIWindowSceneDelegate {
             object: [
                 "scene": scene,
                 "URLContexts": URLContexts,
-            ]
+            ] as [String: Any]
         )
 
         // scene(_:openURLContexts:)


tido64 (approved):
Thanks! I hadn't gotten around to try 14.3 yet. Glad that someone did 😄

tido64:
@fortmarek: Don't worry about the failing commit message linter. We can just change it on merge.

fortmarek:
> Don't worry about the failing commit message linter. We can just change it on merge.

Ok, thanks 👍

PR #1392, opened 2023-04-14
fix(android): fix multiple Kotlin Gradle plugin loads warning

tido64:
### Description

Building the example app currently outputs the following warning:

```
% ./gradlew assembleDebug
Configuration on demand is an incubating feature.

The Kotlin Gradle plugin was loaded multiple times in different subprojects, which is not supported and may break the build.
This might happen in subprojects that apply the Kotlin plugins with the Gradle 'plugins { ... }' DSL if they specify explicit versions, even if the versions are equal.
Please add the Kotlin plugin to the common parent project or the root project, then remove the versions in the subprojects.
If the parent project does not need the plugin, add 'apply false' to the plugin line.
See: https://docs.gradle.org/current/userguide/plugins.html#sec:subprojects_plugins_dsl
The Kotlin plugin was loaded in the following projects: ':app', ':react-native-safe-area-context'
```

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

There should be no warning when running `./gradlew assembleDebug`.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -7,7 +7,7 @@ buildscript {
 plugins {
     id("com.android.application")
     id("com.google.devtools.ksp").version("${kspVersion}")
-    id("org.jetbrains.kotlin.android").version("${kotlinVersion}")
+    id("org.jetbrains.kotlin.android")
 }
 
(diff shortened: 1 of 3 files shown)


kelset (approved):
:shipit:

PR #1398, opened 2023-04-14
fix: bump `react-native-safe-area-context` to fix for xcode 14.3

kelset:
### Description

While testing https://github.com/microsoft/react-native-test-app/pull/1363 I noticed that we needed this bump https://github.com/th3rdwave/react-native-safe-area-context/releases/tag/v4.5.1 or iOS won't build.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

Run 
```
cd example
RCT_NEW_ARCH_ENABLED=1 pod install --project-directory=ios
yarn ios
```

without this bump, it will fail with the "value 12.0" error.

### DIFF
--- a/scripts/android-nightly.patch
+++ b/scripts/android-nightly.patch
@@ -20,7 +20,7 @@ index c9cd14d..f63758d 100644
  import { Colors, Header } from "react-native/Libraries/NewAppScreen";
 @@ -169,7 +169,6 @@ const App = ({ concurrentRoot }) => {
    );
-
+ 
    return (
 -    <SafeAreaProvider>
        <SafeAreaView style={styles.body}>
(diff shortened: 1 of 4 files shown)


scripts/android-nightly.patch
```diff
@@ -20,7 +20,7 @@ index c9cd14d..f63758d 100644
  import { Colors, Header } from "react-native/Libraries/NewAppScreen";
 @@ -169,7 +169,6 @@ const App = ({ concurrentRoot }) => {
    );
- 
+
```
tido64: Does this patch still work without the whitespace? I thought they were intentional?

kelset: not sure tbh, let me revert and just stage the line change

tido64: Looks like it does.

tido64 (approved):
Thanks

PR #1403, opened 2023-04-20
feat: add post_install option to use_test_app

mfazekas:
Some libraries like @rnmapbox/maps requires `post_install` hooks to cocoa pods. Add post_install option to `use_test_app!`

```jsx
use_test_app!(post_install: ->(installer) { $RNMapboxMaps.post_install(installer) })
```

### DIFF
--- a/ios/test_app.rb
+++ b/ios/test_app.rb
@@ -430,6 +430,7 @@ def use_test_app_internal!(target_platform, options)
 
   post_install do |installer|
     react_native_post_install&.call(installer)
+    options[:post_install]&.call(installer)
 
     test_dependencies = {}
     %w[ReactTestAppTests ReactTestAppUITests].each do |target|


ios/test_app.rb
```diff
@@ -429,6 +429,7 @@ def use_test_app_internal!(target_platform, options)
   end
 
   post_install do |installer|
+    options[:post_install]&.call(installer) if options[:post_install]
     react_native_post_install&.call(installer)
```
tido64: Can we reorder this so that the React Native one always runs first? I'm not sure if people rely on it being run first to modify the environment etc.
```suggestion
    react_native_post_install&.call(installer)
    options[:post_install]&.call(installer)
```

Also, it looks like you don't need the trailing `if` here:
```
Error: ios/test_app.rb:432:5: C: [Correctable] Style/SafeNavigation: Use safe navigation (&.) instead of checking if an object exists before calling the method.
    options[:post_install]&.call(installer) if options[:post_install]
```

tido64 (approved):
Thanks for adding this! I left just a nitpick.

ios/test_app.rb
mfazekas: @tido64 thanks - applied

tido64:
A new version will be automatically published when the CI on `trunk` finishes. Thanks again 😄

PR #1434, opened 2023-05-24
fix: autorecalculate the key view loop on the macOS window

Saadnajmi:
### Description

Without this, keyboard accessibility doesn't work in any macOS test app 🤷🏾

### Platforms affected

- [ ] Android
- [ ] iOS
- [x] macOS
- [ ] Windows

### Test plan

Tested in FluentUI React Native. I can now Tab across the controls without hacks.


tido64:
Thanks for fixing accessibility ❤️

PR #1458, opened 2023-06-16
fix(android): fix compatibility with 0.73/nightlies

tido64:
### Description

Starting with 0.73, we need the following:

- Let `@react-native/gradle-plugin` set JVM target
- Gradle 8.1
- Kotlin 1.8

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

```
git apply scripts/disable-safe-area-context.patch
git apply scripts/android-nightly.patch
npm run set-react-version -- nightly
yarn
cd example/android
./gradlew assembleDebug
```

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -10,8 +10,6 @@ plugins {
     id("org.jetbrains.kotlin.android")
 }
 
-checkEnvironment()
-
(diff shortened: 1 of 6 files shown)


kelset (approved):
Screenshot 2023-06-20 at 10 56 23

can confirm that it works - changes make sense.

PR #1467, opened 2023-06-21
test: introduce e2e tests

tido64:
### Description

There are a number of issues we need to address before we can add these to CI. For instance, Android emulators can spend 5 minutes just to boot up, and they require either a Windows or macOS agent. For now, we will only use e2e tests to automate some of the manual testing against newer versions of React Native.

### Platforms affected

- [x] Android
- [x] iOS
- [ ] macOS
- [ ] Windows

### Test plan

Start the tests and follow the on-screen instructions:

```sh
scripts/test-matrix.sh 0.72
```

After the tests are done, there should be 6 screenshots on disk, one for each configuration/platform.

### DIFF
--- a/example/test/specs/wdio.config.js
+++ b/example/test/specs/wdio.config.js
@@ -1,0 +1,40 @@
+// @ts-check
+/** @type {import("webdriverio").RemoteOptions} */
+module.exports = {
+  port: 4723,
+  capabilities: (() => {
+    const args = process.env["TEST_ARGS"]?.toLowerCase() ?? "";
+    const [targetPlatform, ...flags] = args.split(" ");
+
+    const features = {
+      "react:hermes": flags.includes("hermes"),
+      "react:fabric": flags.includes("fabric"),
+      "react:concurrent": flags.includes("concurrent"),
+    };
+
+    switch (targetPlatform) {
+      case "android":
+        return {
+          platformName: "Android",
+          "appium:app": "./android/app/build/outputs/apk/debug/app-debug.apk",
+          "appium:deviceName": "Android GoogleAPI Emulator",
+          "appium:platformVersion": "13.0",
+          "appium:automationName": "UiAutomator2",
+          ...features,
+        };
+      case "ios":
+        return {
+          platformName: "iOS",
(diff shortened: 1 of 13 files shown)


example/test/specs/wdio.config.js
```diff
@@ -0,0 +1,40 @@
...
+        };
+      case "ios":
+        return {
+          platformName: "iOS",
+          "appium:app": "com.microsoft.ReactTestApp",
+          "appium:deviceName": "iPhone 14",
+          "appium:platformVersion": "16.4",
+          "appium:automationName": "XCUITest",
+          ...features,
+        };
```
kelset, line 33: headsup:
for some reason, if I don't have anything open ahead of time for iOS it fails to run the simulator we want and instead picks "the first one available":
Screenshot 2023-07-14 at 11 44 20

which is iPhone SE (3rd generation). So it installs the app there, and then it fails to find it on an iPhone 14 so the test fails:
Screenshot 2023-07-14 at 11 45 56

I'm not sure if there's a way to ensure that it always installs on the right simulator, if not we need to somewhat force it or tell the user to do it ahead of time.

I can't find anything different in the RN core PR https://github.com/facebook/react-native/pull/36267
but maybe I always had the simulators open ahead of time when testing that one...

tido64: Ah, this is because `react-native run-ios` defaults to the first iPhone simulator it finds. We need to add `--simulator 'iPhone 14'` to the `build_and_run` function. I'd rather not hard-code this value tho, so we need to figure out how to centralize this string.

kelset: I think back in the days CLI used to have a way to parse the list of all available simulators and pick the "right one"/current. Maybe we can do something similar. I think just hardcoding it for now for consistency (after all it's hardcoded in the appium settings anyway) would be more than fine.

kelset (approved):

example/test/specs/wdio.config.js
tido64: Hard-coding to `iPhone 14` for now. We can look into this later.

PR #1475, opened 2023-06-27
fix(android): use AGP version provided by Gradle plugin

tido64:
### Description

~~Adapting to the following changes: https://github.com/cortinico/react-native/commit/68fe124fdc0d03d2d44942ba86c90df0b20d81f1~~

- Default to Gradle plugin's recommended version of AGP
- Bump Kotlin to 1.8.22 for latest bug fixes
- Use Moshi 1.15 if Kotlin version is 1.8+

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

These steps will no longer work as is. You'll need to temporarily revert 815bc8b4.

```
npm run clean
git apply scripts/disable-safe-area-context.patch
git apply scripts/android-nightly.patch
npm run set-react-version -- nightly
yarn
cd example/android
./gradlew assembleDebug
```

### Screenshot

Screenshot_1687945153


android/app/src/main/java/com/microsoft/reacttestapp/MainActivity.kt
```diff
@@ -78,7 +78,7 @@ class MainActivity : ReactActivity() {
 
                 useAppRegistry = components.isEmpty()
                 if (useAppRegistry) {
-                    testApp.reactNativeHost.addReactInstanceEventListener {
```
tido64: Seems like this was unintentional. I will revert these for now.

tido64: Upstream fix: https://github.com/facebook/react-native/pull/38088

sweggersen (approved):

PR #1522, opened 2023-07-31
fix(apple): exclude unused classes in single app mode

tido64:
### Description

`UserDefaults` can be misused for fingerprinting and developers are now required to provide a reason for using it. This was announced at WWDC 2023 (see https://developer.apple.com/videos/play/wwdc2023/10060/?time=457). By excluding unused modules when in single app mode, we also conveniently remove all uses of it.

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [ ] Windows

### Test plan

Enable single app mode:

```diff
diff --git a/example/app.json b/example/app.json
index ef69fa7..8e4ddc7 100644
--- a/example/app.json
+++ b/example/app.json
@@ -2,10 +2,12 @@
   "$schema": "https://raw.githubusercontent.com/microsoft/react-native-test-app/trunk/schema.json",
   "name": "Example",
   "displayName": "Example",
+  "singleApp": "example-app",
   "components": [
     {
       "appKey": "Example",
-      "displayName": "App"
+      "displayName": "App",
+      "slug": "example-app"
     },
     {
       "appKey": "Example",
```

Build and run:

```
yarn ios --simulator 'iPhone 14'
```

In a separate terminal, start the dev server:

```
yarn start
```

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -236,7 +236,7 @@ PODS:
   - React-jsinspector (0.71.12)
   - React-logger (0.71.12):
     - glog
-  - react-native-safe-area-context (4.7.0):
+  - react-native-safe-area-context (4.7.1):
     - React-Core
   - React-perflogger (0.71.12)
   - React-RCTActionSheet (0.71.12):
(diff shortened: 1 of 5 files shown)


Saadnajmi (approved):
Hmm... I18nManager in React Native uses user defaults.. I had made a change a while ago to try and remove that use but it caused regressions. I wonder what React native will have to do.

tido64:
> I wonder what React native will have to do.

Prefilling the reason in the template seems like a simple enough workaround?

Issue #111, opened 2020-06-11
Hook up the window from the storyboard to the `AppDelegate`


tido64:
> It seems like a better idea to either hookup the window from the nib/storyboard to the `AppDelegate`; _or_, if the nib/storyboard doesn’t do anything of interest, just remove the window from there and create it programmatically instead.

_Originally posted by @alloy in https://github.com/microsoft/react-native-test-app/pull/104_

tido64:
We will likely migrate to SwiftUI and deprecate `Main.storyboard` entirely. This will require us to bump the macOS deployment target to 11.0+. At the moment, we're stuck on 10.15 so we're looking at 4.0 release at the earliest.

In any case, I did some preliminary investigation and it looks like we no longer need to declare the default menu items. The code below will correctly insert the React menu between View and Window:

```swift
import SwiftUI

@main
struct ReactTestApp: App {
    @State private var rememberLastComponent = false

    var body: some Scene {
        WindowGroup {
            ContentView()
        }
        .commands {
            CommandMenu(LocalizedStringKey(stringLiteral: "React")) {
                Button(LocalizedStringKey(stringLiteral: "Load Embedded JS Bundle")) {
                    // TODO
                }
                Button(LocalizedStringKey(stringLiteral: "Load From Dev Server")) {
                    // TODO
                }
                Toggle(LocalizedStringKey(stringLiteral: "Remember Last Opened Component"), isOn: $rememberLastComponent)
                Divider()
            }
        }
    }
}
```

PR #1527, opened 2023-08-01
fix(macos): fix build break introduced in 2.5.13

tido64:
### Description

Fix build break introduced in #1522

### Platforms affected

- [ ] Android
- [ ] iOS
- [x] macOS
- [ ] Windows

### Test plan

Enable single app mode:

```diff
diff --git a/example/app.json b/example/app.json
index ef69fa7..8e4ddc7 100644
--- a/example/app.json
+++ b/example/app.json
@@ -2,10 +2,12 @@
   "$schema": "https://raw.githubusercontent.com/microsoft/react-native-test-app/trunk/schema.json",
   "name": "Example",
   "displayName": "Example",
+  "singleApp": "example-app",
   "components": [
     {
       "appKey": "Example",
-      "displayName": "App"
+      "displayName": "App",
+      "slug": "example-app"
     },
     {
       "appKey": "Example",
```

Build and run:

```
yarn macos
```

In a separate terminal, start the dev server:

```
yarn start
```

### DIFF
--- a/macos/ReactTestApp/AppDelegate.swift
+++ b/macos/ReactTestApp/AppDelegate.swift
@@ -69,16 +69,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
 
     @IBAction
     func onRememberLastComponentSelected(_ menuItem: NSMenuItem) {
-        switch menuItem.state {
-        case .mixed, .on:
-            Session.shouldRememberLastComponent = false
-            menuItem.state = .off
-        case .off:
-            Session.shouldRememberLastComponent = true
-            menuItem.state = .on
-        default:
-            assertionFailure()
-        }
+        onRememberLastComponentSelectedInternal(menuItem)
     }
 
     // MARK: Private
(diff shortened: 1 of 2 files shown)


macos/ReactTestApp/AppDelegate.swift
```diff
@@ -300,13 +304,11 @@ extension AppDelegate {
...
         #if DEBUG
-        if Session.shouldRememberLastComponent {
-            rememberLastComponentMenuItem.state = .on
-        }
-
         showReactMenu()
         #endif // DEBUG
     }
+
+    private func onRememberLastComponentSelectedInternal(_: NSMenuItem) {}
```
Saadnajmi, line 311: So singleAppMode doesn't need to care about this so it's an empty function? cool :P

Saadnajmi (approved):

macos/ReactTestApp/AppDelegate.swift
tido64: It doesn't make sense since you'll always launch into the app declared in `app.json`. But the Storyboard file needs this to be present ¯\\_(ツ)_/¯

If we recreate whatever the Storyboard does programmatically, we can remove this empty function.

tido64: Somewhat related: https://github.com/microsoft/react-native-test-app/issues/111

Saadnajmi: Somewhat related, visionOS doesn't support storyboards (or was it xibs?) so I think so it would be good to remove uses in general

PR #1563, opened 2023-08-30
fix(init): consolidate `init-test-app` command

tido64:
### Description

Consolidate `init-test-app` command. We currently have two "frontends" for initializing an app. This augments the init script to take arguments, and removes the `@react-native-community/cli` plugin as it wasn't advertised anywhere and was only used on CI.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

```
# Takes the user through the wizard if no arguments are passed
yarn init-test-app

# Takes the user through the wizard, but skips asking for name
yarn init-test-app --name sample

# Skips the wizard, creates an app targeting Android and iOS only
yarn init-test-app --name sample --destination sample -p android -p ios
```


kelset (approved):
:shipit:

PR #1581, opened 2023-09-13
chore: extract config files

tido64:
### Description

- Extracted config files and hid them under `.github/`
- Replaced `node -e 'console.log(...)'` with `node --print '...'`

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

n/a


kelset (approved):
Interesting approach! IIRC the files in the .github folder *might* also be shareable across different repos or something like that... might be worth to keep in mind if we need to share these configs/replicate across multiple repos

PR #1600, opened 2023-09-21
Breaking changes in 3.0

### LINKED ISSUE #1153: Breaking changes in 3.0
PR: https://github.com/microsoft/react-native-test-app/pull/1600
Timeline: Q1 2024

### Breaking changes

- [x] End support for 0.64, 0.65
- [x] Remove workarounds for 0.64, 0.65
  - See [Patches](https://github.com/microsoft/react-native-test-app/wiki/Patches) for full list
- [x] Drop support for Node 14?
  - EOL is [2023-04-30](https://github.com/nodejs/release#release-schedule)
- [x] Drop Flipper support ([RFC](https://github.com/react-native-community/discussions-and-proposals/pull/641))
- [x] Bump requirement for New Architecture

#### iOS

- [x] Bump deployment target to 14.0
- [x] Bump C++ language standard to C++20

#### macOS

- [ ] ~~Bump deployment target to macOS 11~~
  - We should probably stay close to `react-native-macos`
- [x] Bump C++ language standard to C++20

### Post-publish tasks

- [x] Create the `releases/2.x` branch
- [x] Update [dependencies](https://github.com/microsoft/react-native-test-app/wiki#dependencies)
- [x] Move Flipper to a deprecated section
  - [x] [Android](https://github.com/microsoft/react-native-test-app/wiki/Android-Specifics#flipper)
  - [x] [iOS](https://github.com/microsoft/react-native-test-app/wiki/iOS-and-macOS-Specifics#flipper-ios-only)
- [x] Update [Patches](https://github.com/microsoft/react-native-test-app/wiki/Patches)
- [ ] Announcements

tido64:
### Description

This PR collects all breaking changes for the upcoming 3.0 cut.

> **Important**
> To preserve history, this PR must be **rebased** before merging.
> image

Resolves #1153.

#### Blocked by:

- [x] #1730
- [x] #1752
- [x] #1753
- [x] #1755

### Platforms affected

- [x] Android
- [x] iOS
- [x] macOS
- [x] Windows

### Test plan

```
npm run set-react-version 0.66
yarn
# Switch to Node 16
cd example
npm run android
npm run ios

# Switch to Node 18+

# This should throw an exception when it tries to build with new architecture enabled
node scripts/test-matrix.mjs 0.70

# Enabling new architecture should… enable new architecture
node scripts/test-matrix.mjs 0.71
node scripts/test-matrix.mjs 0.73
```


kelset:
comment to keep track of my testing of the branch following the steps highlighted. (finished testing, see my next comment for a quick roundup) 

<details>
<summary>
Testing 0.66 ✅
</summary>


Went ahead, and aside from having to use `YARN_IGNORE_NODE=1` in front of the `yarn` commands (because of Node 16), everything went smoothly. Here's a screenshot of everything working fine:
RNTA 3 0 RN66


</details>

<details>
<summary>Testing 0.70 ✅</summary>

Did a `git clean`, moved to Node LTS (20.10.0), yarn install, then the script. Hit this failure:
```
* What went wrong:
Could not determine the dependencies of task ':app:packageDebug'.
> Could not create task ':app:mergeDebugAssets'.
   > Cannot use @TaskAction annotation on method IncrementalTask.taskAction$gradle_core() because interface org.gradle.api.tasks.incremental.IncrementalTaskInputs is not a valid parameter to an action method.
```

~Should be caused by Gradle & AGP not being compatible. Digging 🦫~
Update: found the problem. Had to manually comment out L27 of `test-matrix.sh` because it was applying the android-nightly patch, hence bumping gradle to 8.x

After addressing that, everything worked out ok as expected:
RNTA 3 0 RN70

And yes, it fails when trying new arch:
RNTA 3 0 RN70 new arch

</details>




<details>
<summary>Testing 0.71 ✅ </summary>

For old arch, everything is working as expected:

RNTA 3 0 RN71


For new arch, iOS is failing with an error that I think we've seen before when working on the C++20 PRs. Might be related to that:
Screenshot 2023-12-13 at 14 48 18

With all the 0.73 stuff fixed, that also fixed the issue that was happening here so now I'm happy to report that the test matrix is also fully working here:

Screenshot 2023-12-18 at 14 54 27

I only had to make an extra bandaid to avoid the android git patch to get applied.

</details>

<details>

<summary>Testing 0.73 ✅</summary>


This one is a bit finicky. We need the android-nightly patch to get gradle to the right version, and Android seems to be working fine: 

Screenshot 2023-12-13 at 15 10 06


but iOS side things get messy already with old arch and Hermes:

Screenshot 2023-12-13 at 15 47 07

need to investigate further, there seems to be something weird going on with pods and build.

**update:** currently having issues with yarn:
Screenshot 2023-12-15 at 12 10 01

I found out that that version gets looked for because the release crew bumped the side packages in [0.73 branch yesterday](https://github.com/facebook/react-native/commit/8c7fc732b719786014e351415f7f424a0280d52e). Because `@react-native/community-cli-plugin` has been bumped with the commit above and `0.73.11` released, now that 12.3 package gets pulled down because of a ^ dependency in `react-native` to `@react-native/community-cli-plugin` itself [here](https://github.com/facebook/react-native/blob/0.73-stable/packages/react-native/package.json#L100).

That said, why is yarn failing, that's what I'm going to dig into next.

**second update:** with 0.73.1 now out properly the issue above has fixed itself.

**side-update:** while working on the above, I noticed that for `@react-native/babel-preset` we were setting the package to a 0.73-nightly instead of `0.73.18` (you can see it in the screenshot). That's because (it turns out) there were some new nightlies done for this package from main branch (because the release crew did not bump to 0.74 on main right away but only like 2 weeks later) *after* the latest stable patch (aforementioned 73.18) was out, and our script was picking latest without checking the tag. So I added a quick bandaid to filter out nightlies, but we need a more appropriate fix that takes care of considering nightlies when working against main branch/nighlies.

**third update**: after the last few commits (and Tommy's fix) now the 0.73 matrix is all completing successfully:
Screenshot 2023-12-18 at 13 25 10


</details>

tido64:
Thanks for testing this ❤️ 

> For new arch, iOS is failing with an error that I think we've seen before when working on the C++20 PRs.

Looks like we missed a file when adding the Folly workaround: https://github.com/microsoft/rnx-kit/pull/2883

kelset:
update on my previous comment: it's all done, everything is working.

BUT, to make it all work seamlessly, aside from Tommy's fix for the C++20 stuff, I had to do two minor bandaids local commits:

* one to make the android patch only apply if we're trying the 0.73 matrix, since it changes Gradle to 8.x. I'm not sure if we should keep it like this or have some more substantial change somewhere else so that when using 0.73 it will also change gradle side files. ([commit](https://github.com/microsoft/react-native-test-app/pull/1600/commits/3a321ff5500d89a80f632e4df4c66d79b0f3f40c))
* the second one is better handling of the fact that we can't assume that the latest version releases of a package is a non-nightly in non-nightly scenarios. More details in the previous comment. Might be worth doing a better implementation so that what I wrote as an inline comment is handled "more properly". ([commit](https://github.com/microsoft/react-native-test-app/pull/1600/commits/4d81c786ed0ecc66b027d85be3938f38797fa3f5))

I think that ideally what we should do next is to go in main branch, address those things proper, then after merging there, revert my local bandaids and instead cherry pick the proper solution commits from main, and then we should be good to go with this.

kelset:
update: my second point has been addressed by https://github.com/microsoft/react-native-test-app/pull/1753

and the first point by https://github.com/microsoft/react-native-test-app/pull/1730

kelset:
one more round of testing before going ahead and merging.

<details>
<summary>
Testing 0.66 ✅
</summary>

It worked:
Screenshot 2024-01-09 at 16 48 11

A few finicky things:
* pod install had to be run manually and was failing because of Folly, it needed a `pod update RCT-Folly --no-repo-update`
* the `npm run ios` command was failing 'cause it wanted to run an iPhone 13, and completely ignored when I tried to pass the option `--simulator="iPhone 15"`. Nothing big, I ran the iOS app from Xcode directly and it worked.
* had to pass the ENV option to make Metro run: `YARN_IGNORE_NODE=1 yarn start --reset-cache`


</details>

 <details>

<summary>
Testing 0.70 ✅
</summary>

Screenshot of it working:
Screenshot 2024-01-09 at 17 55 28


Notes:
* just like last time, had to manually patch locally the WDIO config to bump the iOS version to 17.2
* in order to make it work, I had to add to L133's options `retryDelay: 500,` ([ref](https://nodejs.org/docs/latest-v20.x/api/fs.html#fsrmsyncpath-options)) so that the rm command would work successfully (without it, it would fail with race conditions)
* it, correctly, errors out when trying to do new arch:
Screenshot 2024-01-09 at 17 55 38


</details>

<details>
<summary>
Testing 0.71 ✅
</summary>

Screenshot of it working:
Screenshot 2024-01-10 at 15 03 54


Notes:
* went flawlessly, probably because of the local commits I did for 0.70 (the retryDelay and WDIO changes)

</details>

<details>
<summary>
Testing 0.73 ✅
</summary>

Screenshot of it working:
Screenshot 2024-01-10 at 15 29 28

Notes:
* went flawlessly, probably because of the local commits I did for 0.70 (the retryDelay and WDIO changes)

</details>

kelset (approved):
I did one more [full round of testing](https://github.com/microsoft/react-native-test-app/pull/1600#issuecomment-1881192373), overall I think we are ready to go!

The only minor thing is that tweak to the options for the retryDelay but I'll send that over as a separate small PR.

Let's 🚢  this bad boy!

tido64:
> The only minor thing is that tweak to the options for the retryDelay but I'll send that over as a separate small PR.

I also did a full round. It's strange that I never hit this `retryDelay` issue on 0.70. Regardless, thanks for the thorough testing :heart:

PR #1620, opened 2023-09-26
fix: fix `-h` and `-v` flags not working as intended

tido64:
### Description

The short forms of `--help` and `--version` (`-h` and `-v` respectively) have been broken since #1367.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

Both of these should print the help message:

```
yarn configure-test-app --help
yarn configure-test-app -h
```

These should print the version number:

```
yarn configure-test-app --version
yarn configure-test-app -v
```

### DIFF
--- a/scripts/configure.js
+++ b/scripts/configure.js
@@ -68,6 +68,9 @@ const { parseArgs } = require("../scripts/parseargs");
  */
 const cliPlatformIOS = "@react-native-community/cli-platform-ios";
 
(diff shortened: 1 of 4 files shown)


kelset:
yup, all working: 
Screenshot 2023-09-26 at 15 16 29

kelset (approved):

PR #1621, opened 2023-09-27
test: replace Jest with Node's test runner

tido64:
### Description

Replace Jest with [Node's test runner](https://nodejs.org/api/test.html).

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

n/a


kelset (approved):
kind of massive but it seems it's mostly just github having troubles tracking most of files renames to .mjs

CI is green so I think we can go ahead and deal with issues that might arise as follow up PRs

Saadnajmi:
Question for my own knowledge: why?

tido64:
> Question for my own knowledge: why?

Two primary reasons: 1) it removes a dependency and 2) I had read that it may be faster than Bun in certain cases.

For the latter, I would have to convert all the tests for a proper comparison. Initially, I saw savings after just a couple tests, but it evened out after the migration was done. Our tests are calling a number of external tools like Gradle and rn-cli, which may hit the network in some cases, so the actual times are highly variable on CI. Further improvements are really low on the list. Our Android builds on Windows is currently the worst culprit.

It did remove a significant number of dependencies though, as you can see in #1622.

PR #1623, opened 2023-09-29
fix: ensure correct handling of versions 0.100+

tido64:
### Description

Ensure that we can correctly handle version 0.100 and above. With the current method, version 0.100 would spill over to the major digits i.e.:

- "0.99.0" -> 9900
- "0.100.0" -> 10000
- "1.0.0" -> 10000

### Platforms affected

- [x] Android
- [x] iOS
- [x] macOS
- [x] Windows

### Test plan

Test with 0.72:

```
cd example
yarn android  # or 'ios', 'macos', 'windows'

# In a separate terminal
yarn start
```

Test with 0.64:

```
npm run set-react-version 0.64
yarn
# Switch to Node 16, otherwise Metro will fail to start
cd example
yarn android  # or 'ios', 'macos', 'windows'

# In a separate terminal
yarn start
```

Ideally, we'd test all versions and combinations, but this should cover a big chunk of cases.


scripts/set-react-version.mjs
```diff
@@ -67,7 +67,7 @@ async function checkEnvironment() {
   // See https://nodejs.org/en/blog/release/v18.14.0.
 
   const [major, minor] = process.versions.node.split(".");
-  const nodeVersion = Number.parseInt(major) * 100 + Number.parseInt(minor);
+  const nodeVersion = Number(major) * 100 + Number(minor);
```
kelset, line 70: just silly me asking questions: `.parseInt()` was redundant and that's why you took it off, correct?

android/dependencies.gradle
```diff
@@ -76,11 +38,41 @@ ext {
...
-    targetSdkVersion = reactNativeVersion > 0 && reactNativeVersion < 6500 ? 29 : 33
+    targetSdkVersion = reactNativeVersion > 0 && reactNativeVersion < v(0, 65, 0) ? 29 : 33
+
+    /**
+     * Returns the recommended Gradle plugin version for the specified React Native
+     * version.
+     */
+    def getDefaultGradlePluginVersion = { reactNativeVersion ->
+        // Gradle plugin version can be found in the template:
+        // https://github.com/facebook/react-native/blob/main/packages/react-native/template/android/build.gradle
```
kelset, line 49: Did you mean to link to https://github.com/facebook/react-native/blob/main/packages/react-native/template/android/gradle/wrapper/gradle-wrapper.properties ?

android/dependencies.gradle
```diff
@@ -76,11 +38,41 @@ ext {
...
+        if (reactNativeVersion == 0 || reactNativeVersion >= v(0, 73, 0)) {
+            return ""
```
kelset, line 51: shouldn't this be 8.4? https://github.com/facebook/react-native/blob/main/packages/react-native/template/android/gradle/wrapper/gradle-wrapper.properties#L3

kelset (commented):
left a few comments, maybe I'm wrong

scripts/set-react-version.mjs
tido64: Yeah, you really only need it if you use a different base than 10.

android/dependencies.gradle
tido64: No, we're looking at the Android Gradle Plugin version, specifically `com.android.tools.build:gradle`.

android/dependencies.gradle
tido64: For Gradle 8, you no longer need to specify a version for `com.android.tools.build:gradle`.

kelset (approved):
:shipit:

PR #1633, opened 2023-10-09
fix(apple): disable `deterministic_uuids` starting with 0.73

tido64:
### Description

Disable `deterministic_uuids` starting with 0.73.

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [ ] Windows

### Test plan

```
npm run set-react-version 0.73
# Manually remove react-native-macos from both `package.json` and `example/package.json`
yarn
cd example
rm ios/Podfile.lock
```

### DIFF
--- a/ios/ReactTestApp/Info.plist
+++ b/ios/ReactTestApp/Info.plist
@@ -43,14 +43,8 @@
 	<false/>
 	<key>NSAppTransportSecurity</key>
 	<dict>
-		<key>NSExceptionDomains</key>
-		<dict>
-			<key>localhost</key>
-			<dict>
-				<key>NSExceptionAllowsInsecureHTTPLoads</key>
-				<true/>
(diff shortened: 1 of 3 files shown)


kelset (approved):
:shipit:

Saadnajmi:
Is this just because RN-macOS already did this, or did something change?

tido64:
> Is this just because RN-macOS already did this, or did something change?

Not sure. This line was introduced in [0.68](https://github.com/facebook/react-native/commit/f595a4e681e75aaf737b6582f69855d76a1f33dd), but you really only needed it if you enable Fabric. In 0.73, I'm seeing this even without Fabric so it looks like some files are being generated regardless now.

Saadnajmi:
Ah, it's done upstream too. Interesting.

PR #1650, opened 2023-10-16
chore: override versions in both `dependencies`/`devDependencies`

tido64:
### Description

Make sure `set-react-version.mjs` overrides package versions in both `dependencies` and `devDependencies`.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

```
npm run set-react-version 0.70 -- --core-only
```

Verify that 0.72 specific packages like `@react-native/metro-config` are removed. Since we specified `--core-only`, out-of-tree platform packages should also be removed.

### DIFF
--- a/scripts/set-react-version.mjs
+++ b/scripts/set-react-version.mjs
@@ -13,12 +13,14 @@ import { fileURLToPath } from "node:url";
 import { readJSONFile } from "./helpers.js";
 
 /**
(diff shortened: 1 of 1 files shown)


kelset (approved):
working as expected

:shipit:

PR #1661, opened 2023-10-20
ci(apple): Use macOS 13 XL images

Saadnajmi:
### Description

Use macOS 13 XL images

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [ ] Windows

### Test plan

Ci should pass.. faster

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -137,7 +137,7 @@ jobs:
     timeout-minutes: 60
   ios:
     name: "iOS"
-    runs-on: macos-13-xl
+    runs-on: macos-13-xlarge
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


tido64:
| `macos-13` | `macos-13-xl` | `macos-13-xlarge` |
| :-: | :-: | :-: |
| image | image | image |

🤤

kelset (approved):

tido64:
With the XL images, we are now incurring billable time. I think we need to check whether this is okay:

image

kelset:
oh that's a good shout, let me remove my review so that we don't merge until we have clarity on that

kelset (changes requested):
blocking until we have clarity on billable time

tido64:
Followed up internally and got the message that we're not supposed to be using these. Closing.

PR #1662, opened 2023-10-20
ci: build only affected platforms

tido64:
### Description

Only build platforms that are affected by changes in the PR.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

For each commit you see below, click on the green checkmark on the right side:

image

And verify whether the native bits were built:

image

Note that the "review" jobs are always run in full.

- For the initial commit, no native apps should be built
- For the subsequent commits, the commit message determines what we expect to be built
- For the last commit, all apps should be built

### DIFF
--- a/.github/actions/affected/action.yml
+++ b/.github/actions/affected/action.yml
@@ -1,0 +1,37 @@
+name: Affected Platforms
+description: Determines which platforms are affected by the current changes
+inputs:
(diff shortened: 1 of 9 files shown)


kelset (approved):
🚀

PR #1655, opened 2023-10-18
fix: add Flipper deprecation warning

tido64:
### Description

Flipper has been removed upstream. We will also be removing it from 3.0.

### Platforms affected

- [x] Android
- [x] iOS
- [ ] macOS
- [ ] Windows

### Test plan

Verify that there are no warnings when running:

```
cd example/android
./gradlew clean  # No Flipper warnings here
cd ..
pod install --project-directory=ios  # No Flipper warnings here
```

Enable Flipper:

```diff
diff --git a/example/android/gradle.properties b/example/android/gradle.properties
index a951019..feed9af 100644
--- a/example/android/gradle.properties
+++ b/example/android/gradle.properties
@@ -29,7 +29,7 @@ android.enableJetifier=true

 # Version of Flipper to use with React Native. Default value is whatever React
 # Native defaults to. To disable Flipper, set it to `false`.
-#FLIPPER_VERSION=0.182.0
+FLIPPER_VERSION=0.182.0

 # Enable Fabric at runtime.
 #USE_FABRIC=1
diff --git a/example/ios/Podfile b/example/ios/Podfile
index 245fc9a..be8485f 100644
--- a/example/ios/Podfile
+++ b/example/ios/Podfile
@@ -2,7 +2,7 @@ require_relative '../node_modules/react-native-test-app/test_app'

 workspace 'Example.xcworkspace'

-use_flipper! false unless ENV['ENABLE_FLIPPER']
+use_flipper!

 options = {
   :fabric_enabled => false,
```

With Flipper enabled, we should see warnings:

```
cd android
./gradlew clean  # Flipper deprecation warning
cd ..
pod install --project-directory=ios  # Flipper deprecation warning
```

Finally, verify that `react-native@nightly` builds: https://github.com/microsoft/react-native-test-app/actions/runs/6556025689/job/17805367241

PR #1669, opened 2023-10-23
fix: actually disable Flipper on Android and iOS (both local example app and new generation)

kelset:
### Description

While locally testing https://github.com/microsoft/react-native-test-app/pull/1655 I noticed that I was getting Flipper warnings even when technically Flipper was disabled (since in the `gradle.properties` the line was commented out). See here:

Screenshot 2023-10-23 at 16 14 30

So I read the comment right above it: `To disable Flipper, set it to false.` soooo I did that. And then it actually worked (= the Flipper warnings were not shown).

After the first draft, the PR has been expanded to cover also the iOS use case (for generating new projects we have a custom file `configure.js` that was generating the Podfile instead of copying the one in the example app - which instead happens for Android). 

### Platforms affected

- [x] Android
- [x] iOS
- [ ] macOS
- [ ] Windows

### Test plan

The easiest way is to actually check out https://github.com/microsoft/react-native-test-app/pull/1655, and follow the steps:

```
cd example/android
./gradlew clean  # No Flipper warnings here
```

You will see that the warnings still show if the version stays as commented line, and will only properly disappear the change extracted here is done (meaning when the version is set to `false`).

### DIFF
--- a/example/android/gradle.properties
+++ b/example/android/gradle.properties
@@ -28,8 +28,9 @@ android.useAndroidX=true
 android.enableJetifier=true
 
 # Version of Flipper to use with React Native. Default value is whatever React
-# Native defaults to. To disable Flipper, set it to `false`.
-#FLIPPER_VERSION=0.182.0
+# Native defaults to. To enable Flipper, set the value to `X.Y.Z` (version you want to use).
+# To disable Flipper, set it to `false`.
+FLIPPER_VERSION=false
 
 # Enable Fabric at runtime.
 #USE_FABRIC=1
--- a/scripts/configure.js
+++ b/scripts/configure.js
@@ -643,7 +643,7 @@ const getConfig = (() => {
             Podfile: join(
               `require_relative '${testAppRelPath}/test_app'`,
               "",
-              "use_flipper! false unless ENV['USE_FLIPPER']",
+              "use_flipper! false unless ENV['USE_FLIPPER'] == '1'",
               "",
               `workspace '${name}.xcworkspace'`,
               "",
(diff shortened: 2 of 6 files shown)


tido64 (approved):

example/android/gradle.properties
```diff
@@ -29,7 +29,7 @@ android.enableJetifier=true
 
 # Version of Flipper to use with React Native. Default value is whatever React
 # Native defaults to. To disable Flipper, set it to `false`.
```
tido64: We should probably also change the comment now that Flipper is off by default.

tido64: Also just realized that it's on by default on iOS as well, so we should either disable both or just leave it as is for now?

kelset: whichever you prefer, should we just disable it? I don't think we use it anywhere but not sure.

tido64: Let's disable it on iOS as well:
https://github.com/microsoft/react-native-test-app/blob/d0292b20fed1d4afaa8382a85d65379f7fe3e1a2/ios/test_app.rb#L55-L61

Then do a minor bump. It's a backwards compatible change so it should be fine.

scripts/configure.js
```diff
@@ -643,7 +643,7 @@ const getConfig = (() => {
             Podfile: join(
               `require_relative '${testAppRelPath}/test_app'`,
               "",
-              "use_flipper! false if ENV['USE_FLIPPER'] == '0'",
+              "use_flipper! false unless ENV['USE_FLIPPER']",
```
tido64: Can we make an explicit check in case people are already setting `USE_FLIPPER=0`:
```suggestion
              "use_flipper! false unless ENV['USE_FLIPPER'] == '1'",
```

tido64 (approved):
:shipit:

PR #1679, opened 2023-10-30
fix(ios): replace `LaunchScreen.storyboard` with `UILaunchScreen`

tido64:
### Description

Replaces `LaunchScreen.storyboard` with a much simpler [`UILaunchScreen`](https://developer.apple.com/documentation/bundleresources/information_property_list/uilaunchscreen).

### Platforms affected

- [ ] Android
- [x] iOS
- [ ] macOS
- [ ] Windows

### Test plan

```
cd example
yarn ios

# In a separate terminal
yarn start
```


ios/ReactTestApp/Info.plist
```diff
@@ -65,8 +65,8 @@
 			</array>
 		</dict>
 	</dict>
-	<key>UILaunchStoryboardName</key>
-	<string>LaunchScreen</string>
+	<key>UILaunchScreen</key>
```
tido64, line 68: Turns out this requires iOS 14+ so I retargeted to the 3.0 branch. I will also have to bump iOS minimum version.

kelset (approved):
LGTM - also tested with cocoapods 1.14 and all is working as expected

PR #1686, opened 2023-11-01
chore: restrict use of `export default`

tido64:
### Description

Restrict use of `export default` and convert the `fs` mock to ESM.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

CI should pass.


kelset (approved):
🚢

PR #1715, opened 2023-12-01
chore(macos): treat Swift warnings as errors

tido64:
### Description

Enable `SWIFT_TREAT_WARNINGS_AS_ERRORS` and fix errors.

### Platforms affected

- [ ] Android
- [ ] iOS
- [x] macOS
- [ ] Windows

### Test plan

n/a

### DIFF
--- a/macos/ReactTestApp/AppDelegate.swift
+++ b/macos/ReactTestApp/AppDelegate.swift
@@ -15,6 +15,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     }()
 
     private var manifestChecksum: String?
+    private var contentDidAppearToken: NSObjectProtocol?
 
     func applicationShouldTerminateAfterLastWindowClosed(_: NSApplication) -> Bool {
         true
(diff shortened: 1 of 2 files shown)


Saadnajmi (approved):

macos/ReactTestApp/AppDelegate.swift
```diff
@@ -237,18 +238,19 @@ extension AppDelegate {
...
-                using: { _ in
+                using: { [weak self] _ in
                     #if USE_FABRIC
                     rootView.frame = modalFrame
                     #else
                     (rootView as? RCTRootView)?.contentView.frame = modalFrame
                     #endif
-                    NotificationCenter.default.removeObserver(token!)
+                    if let token = self?.contentDidAppearToken {
+                        NotificationCenter.default.removeObserver(token)
```
Saadnajmi, line 252: Do you need to call removeObservor at all? I thought those were auto removed on de-init?

Saadnajmi: https://developer.apple.com/documentation/foundation/notificationcenter/1413994-removeobserver

> If your app targets iOS 9.0 and later or macOS 10.11 and later, and you used [addObserver(_:selector:name:object:)](https://developer.apple.com/documentation/foundation/notificationcenter/1415360-addobserver), you do not need to unregister the observer. If you forget or are unable to remove the observer, the system cleans up the next time it would have posted to it.

tido64: We're using [addObserver(forName:object:queue:using:)](https://developer.apple.com/documentation/foundation/notificationcenter/1411723-addobserver) though:

> If you used [addObserver(forName:object:queue:using:)](https://developer.apple.com/documentation/foundation/notificationcenter/1411723-addobserver) to create your observer, you should call this method or [removeObserver(_:name:object:)](https://developer.apple.com/documentation/foundation/notificationcenter/1407263-removeobserver) before the system deallocates any object that [addObserver(forName:object:queue:using:)](https://developer.apple.com/documentation/foundation/notificationcenter/1411723-addobserver) specifies.

Or are you saying we should use [addObserver(_:selector:name:object:)](https://developer.apple.com/documentation/foundation/notificationcenter/1415360-addobserver) instead? I think we can do that.

Saadnajmi: If you like, it's not a blocking comment at all or anything

PR #1720, opened 2023-12-05
fix: replace `minimist` with `util.parseArgs`

tido64:
### Description

Now that we require Node 16, we can replace `minimist` with [`util.parseArgs`](https://nodejs.org/api/util.html#utilparseargsconfig).

Blocked by #1600.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

`--help` and `--version` need to be tested manually, otherwise CI should test the rest:

```
yarn configure-test-app --help
yarn configure-test-app -h
yarn configure-test-app --version
yarn configure-test-app -v
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,6 @@
     "chalk": "^4.1.0",
     "cliui": "^8.0.0",
     "fast-xml-parser": "^4.0.0",
-    "minimist": "^1.2.8",
     "prompts": "^2.4.0",
     "semver": "^7.3.5",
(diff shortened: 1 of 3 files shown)


kelset (approved):
Tested locally, all works 👍

Changes LGTM ✅


And GH Copilot is also happy with them 🤖

Screenshot 2024-01-16 at 13 00 08

PR #1761, opened 2024-01-10
test: port `test-e2e.sh` to JS

tido64:
### Description

Port `test-e2e.sh` to JS to allow more complex logic to be written, for instance to avoid installing/updating Appium drivers more than once per session.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

```sh
node scripts/test-matrix.mjs
```

When the run is done, verify the six screenshots under `example/`.


scripts/test-e2e.mjs
```diff
@@ -0,0 +1,153 @@
...
+  if [[ -n $CI ]]; then
+    # Accept all licenses so we can download an Android image
+    yes 2> /dev/null | $sdkmanager --licenses
+    $sdkmanager --install "$android_image"
+
+    # Create an Android emulator and boot it up
+    echo "no" | $avdmanager create avd --package "$android_image" --name Android_E2E
+    $emulator @Android_E2E -delay-adb -partition-size 4096 -no-snapshot -no-audio -no-boot-anim -no-window -gpu swiftshader_indirect &
+  fi
+   */
```
kelset, line 105: do we need this still?

kelset (approved):
Left one comment, LGTM aside from that, tested locally and everything worked as expected:

Screenshot 2024-01-10 at 16 00 59

scripts/test-e2e.mjs
tido64: Yeah, I wanted to keep it around in case we want to try getting this running on CI again. It was a pain to figure out so I'd rather not lose it.

PR #1772, opened 2024-01-15
ci(macos): temporarily disable until publishing is fixed

tido64:
### Description

macOS nightlies constantly fail because `react-native-macos` doesn't publish all the necessary packages:

image

### Platforms affected

- [ ] Android
- [ ] iOS
- [x] macOS
- [ ] Windows

### Test plan

n/a

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -336,6 +336,7 @@ jobs:
   macos:
     name: "macOS"
     runs-on: macos-13
+    if: ${{ github.event_name != 'schedule' }}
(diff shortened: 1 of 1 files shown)


kelset (commented):
let's hope it get fixed soon

kelset (approved):

Saadnajmi:
Oh no! I'll look into what happened in a bit

Saadnajmi:
It seems nightlies indeed have issues. I don't know there's an obvious solution from the first look.

PR #1774, opened 2024-01-16
fix(android): use default app setup for New Arch

tido64:
### Description

Use default app setup for New Arch as it includes a lot of the things that we had to manually implement. Incidentally, this also fixes the crash we had on startup with 0.74 + New Arch.

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

Run the following for 0.71, 0.74:

```
npm run set-react-version 0.71
sed -i '' 's/#newArchEnabled/newArchEnabled/' example/android/gradle.properties
yarn
cd example
yarn android

# In a separate terminal
yarn start
```

### Screenshots

| 0.71 | 0.72 | 0.74 |
| :-: | :-: | :-: |
| Screenshot_1705418147 | Screenshot_1705417852 | Screenshot_1705418519 |


kelset:
Tested, here are the screenshots:






| 0.71 | 0.72 | 0.74 |
| :-: | :-: | :-: |
|Screenshot_1705675500  | Screenshot_1705675952 | Screenshot_1705677330 |

had to jump through a few hoops like the macos problem we are having with their nightlies - but aside from that it all works.

kelset (approved):
:shipit:

PR #1778, opened 2024-01-17
fix(apple): fix build failing on 0.71.12 or lower

tido64:
### Description

`+jsBundleURLForBundleRoot:packagerHost:enableDev:enableMinification:inlineSourceMap:` was first added in 0.71.13

### Platforms affected

- [ ] Android
- [x] iOS
- [ ] macOS
- [ ] Windows

### Test plan

```
npm run set-react-version 0.71
yarn
cd example
yarn ios

# Manually downgrade to 0.71.12 and rebuild
```


kelset (approved):
good stuff 👍

changes make sense and tested locally with 71 and 70 and all works now 🎉

PR #1779, opened 2024-01-17
fix: start using `@rnx-kit/metro-config`

tido64:
### Description

`@rnx-kit/metro-config` is better at handling nested projects and monorepos out-of-box. The intent is to reduce the number of things users have to set up when they generate a new RNTA project.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

CI should pass.


kelset (approved):
I can't wait to try it out, I think it's gonna make things so much easier!

PR #1781, opened 2024-01-18
fix: print selected version when generating projects

tido64:
### Description

Print the selected version when generating projects within an existing project.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

```
% yarn init-test-app --destination test-app --name TestApp -p android -p ios -p macos -p windows
Using 0.73.2 because the current project uses it (use --version to specify another version)
```

### DIFF
--- a/scripts/init.js
+++ b/scripts/init.js
@@ -110,8 +110,20 @@ function getVersion(platforms) {
     return process.argv[index + 1];
   }
 
+  /** @type {(version: string, reason: string) => void} */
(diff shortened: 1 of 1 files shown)


kelset (approved):
Changes LGTM.
Tested locally and I can confirm it works, even used set-react-version to go to 71 and it correctly prints 71 👍

PR #1782, opened 2024-01-18
fix: use correct Gradle version for new projects

tido64:
### Description

Use correct Gradle version for new projects

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

```
yarn init-test-app --destination test-app --name TestApp -p android -p ios -p macos -p windows
cat test-app/android/gradle/wrapper/gradle-wrapper.properties
# This should point to `gradle-8.4-bin.zip`

rm -r test-app
yarn init-test-app --destination test-app --name TestApp -p android -p ios -p macos -p windows --version 0.72
cat test-app/android/gradle/wrapper/gradle-wrapper.properties
# This should point to `gradle-7.6.3-bin.zip`
```

### DIFF
--- a/scripts/configure.js
+++ b/scripts/configure.js
@@ -384,7 +384,8 @@ const getConfig = (() => {
     disableCache = false
   ) => {
     if (disableCache || typeof configuration === "undefined") {
-      const { name, templatePath, testAppPath, flatten, init } = params;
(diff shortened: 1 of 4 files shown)


kelset (approved):
Perfect!

Screenshot 2024-01-18 at 16 11 19

Thanks for the quick turnaround on this!

PR #1784, opened 2024-01-19
fix: only apply config plugins for current platform

tido64:
### Description

Only apply config plugins for current platform.

### Platforms affected

- [x] Android
- [x] iOS
- [ ] macOS
- [ ] Windows

### Test plan

1. Apply changes to https://github.com/sparkfabrik/sparkfabrik-react-native-idfa-aaid/pull/73
2. `npm run android` should not throw an exception
3. `npm run ios` should add `NSUserTrackingUsageDescription` to `node_modules/react-native-test-app/ios/ReactTestApp/Info.plist`

### DIFF
--- a/ios/test_app.rb
+++ b/ios/test_app.rb
@@ -24,7 +24,7 @@ def app_config(project_root)
   [manifest['name'], manifest['displayName'], manifest['version'], manifest['singleApp']]
 end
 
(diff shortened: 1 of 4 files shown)


kelset (approved):
Thanks for the quick fix! it works:
Screenshot 2024-01-19 at 13 49 25

PR #1790, opened 2024-01-22
fix(android): get JS executor name directly via the executor

tido64:
### Description

Get JS executor name directly via the executor. This fixes a startup crash when bridgeless mode is enabled.

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

This changes how we fetch the JS executor name for the string at the bottom right on the home screen:

| 0.66 | 0.73 |
| :-: | :-: |
| Screenshot_1705934910 | Screenshot_1705935048 |


kelset (approved):
:shipit:

PR #1828, opened 2024-02-08
feat: add support for custom app fonts

### LINKED ISSUE #1185: iOS: Add support for fonts
### Proposal

In iOS, adding a custom font requires updating `UIAppFonts` in `Info.plist` https://developer.apple.com/documentation/uikit/text_display_and_fonts/adding_a_custom_font_to_your_app

Looks like it's not supported right now so I propose an additional `fonts` property under platform in the Manifest (app.json)
```json
{
  "ios": {
    "fonts": [
      "EuclidCircularB-Regular.ttf"
      "EuclidCircularB-Bold.ttf",
    ]
  }
}
```

### Implementation Details

The reading of manifest on iOS could be similar to app icons https://github.com/microsoft/react-native-test-app/blob/511d368b91d75647ca51c9cbd7079215c9617e6f/ios/test_app.rb#L242 and then parse, update the plist using `CFPropertyList` (subdep of current gems). Here's some code (thanks CodePilot), tested in example and it works.

```ruby
require("cfpropertylist")

# ...

def set_fonts_in_plist!(project_root, target_platform, destination)
  fonts = platform_config('fonts', project_root, target_platform)
  return if fonts.nil? || fonts.empty?

  info_plist = File.join(destination, 'ReactTestApp', 'Info.plist')

  plist = CFPropertyList::List.new(file: info_plist)
  data = CFPropertyList.native_types(plist.value)

  data['UIAppFonts'] = []
  fonts.each do |font|
    data['UIAppFonts'] << File.basename(font)
  end

  plist.value = CFPropertyList.guess(data)
  File.write(info_plist, plist.to_str(CFPropertyList::List::FORMAT_XML))
end

# ...
generate_assets_catalog!(project_root, target_platform, destination) 
set_fonts_in_plist!(project_root, target_platform, destination)
```

Also worth adding the font files automatically to resources, otherwise they need to be duplicated in `resources.ios[]`.

### Code of Conduct

### LINKED ISSUE #1506: iOS/macOS: add ability to customize `Info.plist`
### Proposal

Users will need the ability to customize `Info.plist` for their apps. The only way to do that today is to write a [Config Plugin](https://github.com/microsoft/react-native-test-app/wiki/Config-Plugins), e.g.:

```js
const { withInfoPlist } = require("@expo/config-plugins");

module.exports = (config, { description }) => {
  return withInfoPlist(config, (config) => {
    if (!config.ios) {
      config.ios = {};
    }
    if (!config.ios.infoPlist) {
      config.ios.infoPlist = {};
    }

    config.ios.infoPlist["NSPhotoLibraryUsageDescription"] = description;
    return config;
  });
};
```

### Alternatives

There are several ways we can implement this. I still don't have an idea of what would be ideal here.

#### Custom path to `Info.plist`

One option is to let users specify a path to their own `Info.plist`. This means that would have to modify the path in `project.pbxproj` on `pod install`:

https://github.com/microsoft/react-native-test-app/blob/fafd197b184dfdb3b29ecdf548f38d46ac8fd76c/ios/ReactTestApp.xcodeproj/project.pbxproj#L64

#### `app.json`

Another option is to specify fields in `app.json` that we later write to `Info.plist` when generating the project. Today, we are symlinking the entire `ReactTestApp` folder. We will need to split this out somehow.

https://github.com/microsoft/react-native-test-app/blob/fafd197b184dfdb3b29ecdf548f38d46ac8fd76c/ios/test_app.rb#L231-L234

### Implementation Details

n/a

### Additional Context

_No response_

### Code of Conduct

### LINKED ISSUE #1813: iOS / Android Assets
### Proposal

I would like to add assets to native iOS and Android in test app like fonts, mp3 files etc. 

### Alternatives

https://github.com/unimonkiez/react-native-asset

### Implementation Details

I'm using https://github.com/unimonkiez/react-native-asset to inject and link the assets in iOS and Android. But I want to use this library inside your test app. Can you provide an alternate to this on how to inject fonts and link it on both iOS and Android

### Additional Context

_No response_

### Code of Conduct

tido64:
### Description

Add support for custom app fonts.

Resolves #1506.
Resolves #1185.
Resolves #1813.

### Platforms affected

- [x] Android
- [x] iOS
- [x] macOS
- [ ] Windows

### Test plan

1. Download a random font and put it somewhere in the example folder (in my case, I put mine in `example/dist`)
2. Add the font to resources in `app.json` (an entry for every target platform):
    ```diff
    diff --git a/example/app.json b/example/app.json
    index ef69fa7..c5a1910 100644
    --- a/example/app.json
    +++ b/example/app.json
    @@ -16,15 +16,18 @@
       "resources": {
         "android": [
           "dist/res",
    -      "dist/main.android.jsbundle"
    +      "dist/main.android.jsbundle",
    +      "dist/MyFont.ttf"
         ],
         "ios": [
           "dist/assets",
    -      "dist/main.ios.jsbundle"
    +      "dist/main.ios.jsbundle",
    +      "dist/MyFont.ttf"
         ],
         "macos": [
           "dist/assets",
    -      "dist/main.macos.jsbundle"
    +      "dist/main.macos.jsbundle",
    +      "dist/MyFont.ttf"
         ],
         "windows": [
           "dist/assets",
    ```
3. Use the font:
    ```diff
    diff --git a/example/App.tsx b/example/App.tsx
    index c94affc..93821a6 100644
    --- a/example/App.tsx
    +++ b/example/App.tsx
    @@ -136,6 +136,7 @@ function useStyles() {
           groupItemLabel: {
             color: isDarkMode ? Colors.white : Colors.black,
             flex: 1,
    +        fontFamily: "MyFont",
             fontSize,
             marginVertical: 12,
           },
    ```

Build the example app as usual. Remember that every time you make changes to `app.json`, you need to re-run `pod install`.

**Verify that fonts are copied/registered:**
- Android: Your font(s) should be found under `example/android/app/build/generated/rncli/src/main/assets/fonts`
- iOS: Your font(s) should be registered in `example/node_modules/.generated/ios/Info.plist`
- macOS: Your font(s) should be registered in `example/node_modules/.generated/macos/Info.plist`

### Screenshots

| Android | iOS | macOS |
| :-: | :-: | :-: |
| Screenshot_1707395932 | Simulator Screenshot - iPhone 15 Pro - 2024-02-08 at 11 29 22 | Screenshot 2024-02-08 at 11 38 34 |

### DIFF
--- a/test-app.gradle
+++ b/test-app.gradle
@@ -143,13 +143,7 @@ ext.applyTestAppModule = { Project project ->
     preBuild.dependsOn(tasks.register("copyFonts", Copy) {
         androidResourceFiles.each {
             if (isFontFile(it)) {
-                from(it) {
-                    // File-based resource names must contain only lowercase
-                    // a-z, 0-9, or underscore
-                    rename {
-                        it.toLowerCase()
-                    }
-                }
+                from(it)
             }
         }
 
--- a/test-app.gradle
+++ b/test-app.gradle
@@ -16,7 +16,7 @@ private static void applyConfigPlugins(String testAppDir, File rootDir) {
 
 private static boolean isFontFile(File file) {
     // https://github.com/facebook/react-native/blob/3dfedbc1aec18a4255e126fde96d5dc7b1271ea7/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/common/assets/ReactFontManager.java#L28
-    return file.name.endsWith(".ttf") || file.name.endsWith(".otf")
+    return [".otf", ".ttf"].any { file.name.endsWith(it) }
 }
 
 def testAppDir = buildscript.sourceFile.getParent()
--- a/test-app.gradle
+++ b/test-app.gradle
@@ -14,6 +14,11 @@ private static void applyConfigPlugins(String testAppDir, File rootDir) {
     }
 }
 
+private static boolean isFontFile(File file) {
+    // https://github.com/facebook/react-native/blob/3dfedbc1aec18a4255e126fde96d5dc7b1271ea7/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/common/assets/ReactFontManager.java#L28
+    return file.name.endsWith(".ttf") || file.name.endsWith(".otf")
+}
+
 def testAppDir = buildscript.sourceFile.getParent()
 apply(from: "${testAppDir}/android/test-app-util.gradle")
 
@@ -93,15 +98,37 @@ ext.applyTestAppModule = { Project project ->
     def generatedResDir = file("${buildDir}/generated/rncli/src/main/res/")
     generatedResDir.mkdirs()
 
+    // https://github.com/facebook/react-native/blob/3dfedbc1aec18a4255e126fde96d5dc7b1271ea7/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/common/assets/ReactFontManager.java#L25
+    def generatedFontsDir = file("${generatedAssetsDir}/fonts")
+    generatedFontsDir.mkdirs()
+
     def generatedRawDir = file("${generatedResDir}/raw")
     generatedRawDir.mkdirs()
 
     preBuild.dependsOn(tasks.register("copyAssets", Copy) {
         androidResourceFiles.each {
-            from(it)
+            if (!isFontFile(it)) {
+                from(it)
+            }
         }
 
         into(generatedAssetsDir)
+    })
+
+    preBuild.dependsOn(tasks.register("copyFonts", Copy) {
+        androidResourceFiles.each {
+            if (isFontFile(it)) {
+                from(it) {
+                    // File-based resource names must contain only lowercase
+                    // a-z, 0-9, or underscore
+                    rename {
+                        it.toLowerCase()
+                    }
+                }
+            }
+        }
+
+        into(generatedFontsDir)
     })
 
     preBuild.dependsOn(tasks.register("copyResources", Copy) {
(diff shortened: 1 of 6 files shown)


kelset:
Testing:

iOS ✅

Simulator Screenshot - iPhone 15 - 2024-02-12 at 11 18 32

macOS ✅

Screenshot 2024-02-12 at 11 19 07


Android ✅

Screenshot_1707736715


~~The one thing I noticed though is that for Android the font file copied over in `example/android/app/build/generated/rncli/src/main/assets/fonts` was uncapitalized, `pacifico-regular` instead of `Pacifico-Regular`.~~  fixed with latest commit 👍

test-app.gradle
```diff
@@ -93,17 +123,55 @@ ext.applyTestAppModule = { Project project ->
...
     })
 
+    preBuild.dependsOn(tasks.register("copyFonts", Copy) {
+        androidResourceFiles.each {
+            if (isFontFile(it)) {
+                from(it) {
+                    // File-based resource names must contain only lowercase
+                    // a-z, 0-9, or underscore
+                    rename {
+                        it.toLowerCase()
```
kelset: (following up from my testing comment - if this is something we can't work around, we should have in the docs somewhere explaining that they will need to keep it in mind when importing a font, and maybe provide a snippet for how to handle different platforms in the fonts styling)

kelset (approved):
LGTM and we can :shipit: 

the lowercase thing for Android can be handled as a follow up if needed

test-app.gradle
kelset: fixed by latest commit 👍

PR #1827, opened 2024-02-08
chore(set-react-version): remove stale resolutions

tido64:
### Description

Nightlies have been failing because it's trying to use `@react-native/babel-preset` 0.73 instead of nightly:

```
error node_modules/react-native/src/private/specs/components/DebuggingOverlayNativeComponent.js: /~/node_modules/react-native/src/private/specs/components/DebuggingOverlayNativeComponent.js: Unsupported param type for method "highlightTraceUpdates", param "updates". Found $ReadOnlyArray.
Error: /~/node_modules/react-native/src/private/specs/components/DebuggingOverlayNativeComponent.js: Unsupported param type for method "highlightTraceUpdates", param "updates". Found $ReadOnlyArray
```

Build log: https://github.com/microsoft/react-native-test-app/actions/runs/7824683491

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

Run:

```
npm run set-react-version nightly
```

Verify that React Native related resolutions are removed from `package.json`:

```diff
     "@expo/json-file/@babel/code-frame": "^7.10.4",
     "@microsoft/eslint-plugin-sdl/eslint-plugin-react": "^7.33.0",
     "@pnpm/network.ca-file/graceful-fs": "~4.2.10",
-    "@react-native-community/cli": "^12.3.0",
-    "@react-native-community/cli-platform-android": "^12.3.0",
-    "@react-native-community/cli-platform-ios": "^12.3.0",
-    "@react-native/babel-preset": "^0.73.0",
-    "@react-native/community-cli-plugin/@react-native-community/cli-server-api": "^12.3.0",
-    "@react-native/community-cli-plugin/@react-native-community/cli-tools": "^12.3.0",
     "@semantic-release/npm/npm": "link:./example",
     "appium/@types/semver": "^7.5.6",
     "appium/yaml": "^2.3.4",
```

Verify that we are able to bundle:

```
cd example
yarn build:ios
```

PR #1829, opened 2024-02-08
fix(ios): fix crash on startup on nightlies

tido64:
### Description

An assertion was introduced, causing a crash on startup: https://github.com/facebook/react-native/blob/50bd9e367bb8a0c18be43f751a2fc95908df1ff1/packages/react-native/React/CxxBridge/RCTCxxBridge.mm#L439

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [ ] Windows

### Test plan

```
npm run set-react-version nightly
yarn
cd example
yarn ios

# In a separate terminal
yarn start
```

### DIFF
--- a/ios/ReactTestApp/ContentView.swift
+++ b/ios/ReactTestApp/ContentView.swift
@@ -111,7 +111,9 @@ final class ContentViewController: UITableViewController {
         onComponentsRegistered(components, checksum: checksum)
 
         let bundleRoot = manifest.bundleRoot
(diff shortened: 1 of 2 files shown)


kelset (approved):
tested together with https://github.com/microsoft/react-native-test-app/pull/1827 I was able to verify that the fixes work 👍

Simulator Screenshot - iPhone 15 - 2024-02-08 at 15 14 33

tido64 (commented):
Need to make the equivalent change on the macOS side as well.

PR #1837, opened 2024-02-12
fix(apple): enable `concurrentRoot` with New Architecture

tido64:
### Description

Having `concurrentRoot` disabled when Fabric is enabled is not recommended: https://github.com/facebook/react-native/commit/7eaabfb174b14a30c30c7017195e8110348e5f44

As of 0.74, it won't be possible to opt-out: https://github.com/facebook/react-native/commit/30d186c3683228d4fb7a42f804eb2fdfa7c8ac03

On Android, this is already the case.

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [ ] Windows

### Test plan

1. Build:
    ```
    cd example
    RCT_NEW_ARCH_ENABLED=1 pod install --project-directory=ios
    yarn ios --simulator 'iPhone 15 Pro'

    # In a separate terminal
    yarn start
    ```
2. Verify that `Concurrent React` is **on**
3. Opt-out:
    ```diff
    diff --git a/example/app.json b/example/app.json
    index ef69fa7..0ea1dae 100644
    --- a/example/app.json
    +++ b/example/app.json
    @@ -5,7 +5,10 @@
       "components": [
         {
           "appKey": "Example",
    -      "displayName": "App"
    +      "displayName": "App",
    +      "initialProperties": {
    +        "concurrentRoot": false
    +      }
         },
         {
           "appKey": "Example",
    ```
4. Rebuild the app: `yarn ios --simulator 'iPhone 15 Pro'`
5. Verify that `Concurrent React` is **off**

### Screenshots

| Concurrent React enabled | Opted-out |
| :-: | :-: |
| image | image |

PR #1840, opened 2024-02-13
chore(set-react-version): apply patches while setting version

tido64:
### Description

Apply patches while setting version.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

1. Verify that `gradle-wrapper.properties` gets patched: `npm run set-react-version 0.72`
2. Verify that Jetifier gets disabled (`example/android/gradle.properties`): `npm run set-react-version nightly`


scripts/set-react-version.mjs
```diff
@@ -97,6 +97,20 @@ async function checkEnvironment() {
   return true;
 }
 
+/**
+ * Disables [Jetifier](https://developer.android.com/tools/jetifier).
```
kelset, line 101: might be worth adding an extra line to reference why from 74 onward we don't want to use jetifier anymore

scripts/set-react-version.mjs
```diff
@@ -417,18 +448,24 @@ export function setReactVersion(version, coreOnly) {
...
+  } else {
+    setReactVersion(version, process.argv.includes("--core-only")).then(() => {
+      const numVersion =
+        version === "nightly"
+          ? Number.MAX_SAFE_INTEGER
+          : toVersionNumber(version);
+      if (numVersion >= v(0, 74, 0)) {
+        disableJetifier();
+      } else if (numVersion < v(0, 73, 0)) {
+        setGradleVersion("7.6.3");
```
kelset, line 471: based on my testing for the other PR -> https://github.com/microsoft/react-native-test-app/pull/1837#issuecomment-1941437582

we still have the webstorage webapi problem when downgrading and testing Android; do we want/need to have a patch similar to the one we had for safearea to apply here? we can leave it for a follow up PR if yes

kelset (approved):
tested locally, all works. Add a couple of comments but they are low prior/not blocking 👍

scripts/set-react-version.mjs
tido64: We should fix this in a separate PR. I'm prepping some fixes in `web-storage`.

PR #1843, opened 2024-02-14
chore: `@react-native-webapis/web-storage` is not compatible with 0.71

tido64:
### Description

`set-react-version` should remove `@react-native-webapis/web-storage` since it's not compatible with 0.71.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

`@react-native-webapis/web-storage` should be removed when switching to a version lower than 0.72:

```
git checkout .
npm run set-react-version 0.71 -- --core-only
```

Verify that `@react-native-webapis/web-storage` is removed from `example/package.json`.

```
git checkout .
npm run set-react-version 0.72 -- --core-only
```

Verify that `@react-native-webapis/web-storage` is NOT removed from `example/package.json`.

### DIFF
--- a/android/test-app-util.gradle
+++ b/android/test-app-util.gradle
@@ -252,8 +252,8 @@ ext.isBridgelessEnabled = { project, isNewArchEnabled ->
             def isSupported = version == 0 || version >= v(0, 73, 0)
             if (!isSupported) {
                 logger.warn([
-                    "react-native 0.73 or greater is required for Bridgeless",
-                    "Mode — disable `bridgelessEnabled` in your",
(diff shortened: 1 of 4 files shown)


kelset:
Tested 71 on both arch(s) just to be sure:

isolated

isolated

kelset (approved):
tested, all works as expect.
I didn't realize ('cause im dumb) that because it's "pure web code" in the example app, it will gracefully fail even if we don't have the library in the deps/built (which would have happened with ex. safeareacontext)

PR #1847, opened 2024-02-16
fix(apple): fix `pod install` failing if no resources are declared

tido64:
### Description

Fix `pod install` failing if no resources are declared:

```
% pod install --project-directory=ios

[!] Invalid `Podfile` file: undefined method `each' for nil.
```

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [ ] Windows

### Test plan

1. Remove the `resources` field from `example/app.json`
2. Run `pod install` in `example/ios`


kelset (approved):
Good catch 👍

Saadnajmi:
Thanks! I'll update my PR

PR #1848, opened 2024-02-17
fix(android): build from source with `react.buildFromSource`

tido64:
### Description

To build React Native from source, add `react.buildFromSource=true` to `gradle.properties`.

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

```
npm run set-react-version nightly
yarn clean
yarn
cd example
sed -i '' 's/#react.buildFromSource/react.buildFromSource/' android/gradle.properties
yarn android
```

If you hit a build error, see https://github.com/microsoft/react-native-test-app/wiki/Building-React-Native-From-Source


kelset (approved):
yup, can confirm it working. The code changes LGTM

I'm just a bit concerned by the fact that I needed to get NDK 26 (following the wiki) to get it to build successfully. Is this something we need to flag up upstream? they are planning to cut the 74 branch soon

PR #1856, opened 2024-02-20
chore: lift hoisting limits

tido64:
### Description

Re-enable hoisting so we can spot codegen/autolinking issues. This isn't perfect since everything will be hoisted to the root project and there is currently no way to exclude certain packages like in Yarn Classic. The only difference now is that `react-native` will no longer get installed inside the example project folder.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

n/a


kelset (approved):
hoist the flag 🏴‍☠️

PR #1870, opened 2024-02-23
fix: convert scripts to ESM

tido64:
### Description

Convert scripts to ESM. Not all scripts can converted today because we consume them from CJS.

`configure.js` can be converted but will require some refactoring work in a separate PR: https://github.com/microsoft/react-native-test-app/pull/1871

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

CI should pass.


kelset (approved):
LGTM

PR #1877, opened 2024-02-27
ci(windows): disable NuGet setup since we build from source

tido64:
### Description

Disable NuGet setup since we build from source.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [x] Windows

### Test plan

n/a

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -496,7 +496,8 @@ jobs:
         uses: ./.github/actions/affected
       - name: Set up NuGet sources
         # Temporarily disabled since builds using NuGet artifacts randomly fail
-        # with architecture mismatch error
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -495,7 +495,9 @@ jobs:
         id: affected
         uses: ./.github/actions/affected
       - name: Set up NuGet sources
-        if: ${{ steps.affected.outputs.windows != '' && github.event_name == 'schedule' }}
+        # Temporarily disabled since builds using NuGet artifacts randomly fail
```
kelset, line 498: since you mention "temporarily", should we track this with an issue somewhere and add a reference here?

kelset (approved):

PR #1889, opened 2024-03-06
chore: replace `chalk` with `yoctocolors`

tido64:
### Description

Replace `chalk` with `yoctocolors` because it's simpler/smaller than most alternatives.

See benchmarks in their README: https://github.com/sindresorhus/yoctocolors?tab=readme-ov-file#benchmark

[picocolors](https://github.com/alexeyraspopov/picocolors) also has benchmarks (still in PR).

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] Windows

### Test plan

CI should pass.


kelset (approved):
:shipit:

PR #1896, opened 2024-03-11
refactor: Share more xcconfigs across apple platforms

Saadnajmi:
### Description

Refactor the xcconfigs so that the iOS specific settings are pushed to new iOS specific files, and the base common/debug/release can be shared across platforms

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [ ] Windows

### Test plan

CI should pass


Saadnajmi:
Seems there's a legitimate macOS error, which is the one I was worried about if I refactored the visionOS ifdefs:
> Error: could not find included file '../ReactTestAppShared/ReactTestApp.common.xcconfig' in search paths

If you make a template app with just the macOS platform, I was afraid that file may be missing. I'll look into it

Saadnajmi:
I'll probably just close this PR. Not sure when I'll work on it next / how to get past the error I ran into.

tido64:
> I'll probably just close this PR. Not sure when I'll work on it next / how to get past the error I ran into.

Let's revisit this later. Thanks for working on it 😄

PR #1903, opened 2024-03-14
fix: resolve `react-native-test-app` path dynamically

tido64:
### Description

When generating new projects, we hardcode the path to RNTA. This does not always work, especially in monorepos.

### Platforms affected

- [x] Android
- [x] iOS
- [x] macOS
- [x] visionOS
- [ ] Windows

### Test plan

CI should pass.


kelset (approved):
love to see it 👍

Saadnajmi:
Hmm, this may simplify the template app in FURN. I'll take a look when the dependabot PR comes through

PR #1908, opened 2024-03-15
fix(apple): enable bridgeless by default with New Arch in 0.74

tido64:
### Description

Enable bridgeless by default with New Arch in 0.74

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [x] visionOS
- [ ] Windows

### Test plan

n/a


kelset (approved):
makes sense, let's ship it 👍

PR #1919, opened 2024-03-19
fix: replace `mkdirp` with an inline script

tido64:
### Description

For newly generated projects, we can drop `mkdirp` for an inline script.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

n/a


kelset (approved):
LGTM

:shipit:

PR #1928, opened 2024-03-22
fix(apple): respect `USE_HERMES` env var

tido64:
### Description

`USE_HERMES` as added in 0.71

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [x] visionOS
- [ ] Windows

### Test plan

```
cd example
USE_HERMES=1 pod install --project-directory=ios
```

Verify that the output includes `React-hermes`.


kelset (approved):
🚢 

thanks for the quick turnaround!

Issue #1932, opened 2024-03-25
Testing against multiple RN versions


johnf:
### What happened?

Apologies in advance; this is more a question than a bug report.

Should I be able to use this project to create a matrix test in GitHub actions to test my react native library against multiple versions of React Native?

I have the basics of this set up https://github.com/oblador/react-native-vector-icons/tree/monorepo

I'm using detox ti run e2e tests.

This works perfectly with RN 0.73

My intention was to then setup a matrix where I would
* yarn add react-native-test-app@0.72-stable (0.71-stable etc)
* yarn android
* detox ...

As I'm trying to get this working on past versions of RN I'm hitting all sorts of problems like
* Wrong gradle version
* Wrong java version
* Kotlin mismatches

I want to check if what I'm trying to do is feasible before filing more bugs around the issues I'm seeing.

### Version

3.3.6

### What platforms are you seeing this issue on?

- [X] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### System Information

```true
N/A at this stage
```

### Steps to Reproduce

See https://github.com/oblador/react-native-vector-icons/tree/monorepo

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

johnf:
I found one issue that I fixed in #1934 because I'm running a monorepo.

But I think that's only part of the puzzle.

I noticed `test:matrix` in RNTA makes a number of other changes to packages.json, going to look into that shortly

kelset:
Hey @johnf - so, first off thanks for asking!

This is something that we have been thinking on how to make more easily accessible for a while - in particular for situations like BabylonReactNative (they are still migrating to it: https://github.com/BabylonJS/BabylonReactNative/pull/641, but they need to test, similar to you, different RN versions).

As you already found out, we do have a few scripts in this repo to make the process of switching versions of RN for our local test bench swifter:
* https://github.com/microsoft/react-native-test-app/blob/trunk/scripts/test-e2e.mjs
* https://github.com/microsoft/react-native-test-app/blob/trunk/scripts/test-matrix.mjs
* https://github.com/microsoft/react-native-test-app/blob/trunk/scripts/set-react-version.mjs

So maybe an alternative is to tweak those scripts so that they can be more generalised and exported along with RNTA, so that you can just invoke them in a RNTA-based test app.

[...]

PR #1934, opened 2024-03-25
fix(android) Be more explicit about which gradle plugin to include

johnf:
The previous approach could fail in a monorepo where both modules could exist in a hoisted node_modules

### Description

I discovered an issue with selecting which version of the react native module to include.

The current code checks if `@react-native/gradle-plugin` exists (RN >= 0.72) and usesit if it does.

This logic fails in a monorepo situation where both libraries could exist for different packages.

This change explicitly checks the react native version and loas the appropriate library.

This is on the path to helping me solve #1932

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

I'm currently testing this against my code base. Easiest way to reproduce

* Create a yarn workspace
* Create to react-native-test-apps inside
* One on 0.73 and on one 0.70
* Make sure you can build both

### DIFF
--- a/test-app.gradle
+++ b/test-app.gradle
@@ -67,9 +67,12 @@ ext.applyTestAppSettings = { DefaultSettings settings ->
     settings.project(":support")
         .projectDir = file("${testAppDir}/android/support")
 
+    def reactNativeVersion = getPackageVersionNumber("react-native", rootDir)
+
     def reactNativeGradlePlugin =
-        findNodeModulesPath("@react-native/gradle-plugin", reactNativeDir)  // >= 0.72
-            ?: findNodeModulesPath("react-native-gradle-plugin", reactNativeDir)  // < 0.72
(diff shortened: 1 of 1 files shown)


kelset:
hey @johnf one quick headsup, you are missing the `:` so the commit linter fails

kelset (approved):
aside from the commit lint issue, the change is solid and respect existing approch, LGTM 👍

one small nit is that I'm not sure we'd recommend having multiple RNTA apps in the same codebase on different RN versions, but we can explore more that conversation here (https://github.com/microsoft/react-native-test-app/issues/1932)

johnf:
fixed

tido64 (approved):

tido64:
@johnf, this is a good fix. Thanks ❤️

PR #1935, opened 2024-03-25
fix(android): automatically configure Gradle wrapper

tido64:
### Description

Automatically configure Gradle wrapper. To disable this behavior, set `RNTA_CONFIGURE_GRADLE_WRAPPER=0`.

Basically, this is what the script does:

- If 0.74 or greater (including nightlies), set Gradle version to 8.6 if the current version is lower
- If 0.73, set Gradle version to 8.3 if the current version is lower
- If 0.72, set Gradle version to 8.1.1 if the current version is lower
- Else, set Gradle version to 7.6.4 if the current version is <7.5.1 or >=8.0.0

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

1. Switch to 0.71: `npm run set-react-version 0.71 -- --core-only`
2. Verify that `example/android/gradle/wrapper/gradle-wrapper.properties` is unmodified
3. Install npm dependencies: `yarn`
4. Build the Android app: `cd example && yarn android`
5. In a separate terminal, verify that the Gradle version changed to 7.6.4 (in `example/android/gradle/wrapper/gradle-wrapper.properties`)
6. Switch to 0.73: `npm run set-react-version 0.73 -- --core-only`
7. Verify that the Gradle version is still 7.6.4
8. Install npm dependencies: `yarn`
9. Build the Android app: `cd example && yarn android`
10. In a separate terminal, verify that the Gradle version changed to 8.3
11. Switch to 0.72: `npm run set-react-version 0.72 -- --core-only`
12. Verify that the Gradle version is still 8.3
13. Install npm dependencies: `yarn`
14. Build the Android app: `cd example && yarn android`
15. In a separate terminal, verify that the Gradle version is still 8.3
16. Switch to 0.71: `npm run set-react-version 0.71 -- --core-only`
17. Verify that the Gradle version is still 8.3
18. Install npm dependencies: `yarn`
19. Build the Android app but disable the behavior: `cd example && RNTA_CONFIGURE_GRADLE_WRAPPER=0 yarn android`
20. Verify that the Gradle version is still 8.3

### DIFF
--- a/scripts/configure-projects.js
+++ b/scripts/configure-projects.js
@@ -7,6 +7,7 @@
  */
 const nodefs = require("node:fs");
 const path = require("node:path");
+const tty = require("node:tty");
 const {
   findNearest,
   getPackageVersion,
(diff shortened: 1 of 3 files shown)


scripts/configure-projects.js
```diff
@@ -42,6 +43,72 @@ const cliPlatformIOSVersion = (() => {
...
+        ? "\u001B[33m\u001B[1mwarn\u001B[22m\u001B[39m"
+        : "warn";
+      console.warn(tag, `Setting Gradle version ${gradleVersion}`);
+      fs.writeFileSync(
+        gradleWrapperProperties,
+        props.replace(re, `gradle-${gradleVersion}-bin.zip`)
+      );
+    }
+  } catch (_) {
+    // ignore
```
kelset: a small nit: it might be worth to output some warn here to let the user know that something went weird when trying to verify that the right version of Gradle is used in this project. I feel that this might help the user in debugging issues later down the line if ex. the can't get the app to run; we might be pointing them to check out the gradle side of the config and help them.

kelset (approved):
Good stuff!

Tested the whole flow locally and it works great.

Let's :shipit:

PR #1961, opened 2024-04-09
fix(init): use npm public registry api instead of cli

tido64:
### Description

Use npm public registry api instead of spawning separate `npm` processes.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

```
% yarn init-test-app --destination pr1961 --name TestApp -p android -p ios -p macos -p windows
Using 0.73.6 because the current project uses it (use --version to specify another version)

% yarn init-test-app --destination pr1961 --name TestApp -p android -p ios -p macos -p windows --version 0.74
Downloading react-native-0.74.0-rc.7.tgz...
[!] react-native-macos@0.74 cannot be added because it does not exist or is unsupported
[!] react-native-windows@0.74 cannot be added because it does not exist or is unsupported

% yarn init-test-app --destination pr1961 --name TestApp -p android -p ios -p macos -p windows --version 0.74.0-rc.6
Downloading react-native-0.74.0-rc.6.tgz...
[!] react-native-macos@0.74 cannot be added because it does not exist or is unsupported
[!] react-native-windows@0.74 cannot be added because it does not exist or is unsupported

% cd ..
% ./react-native-test-app/scripts/init.mjs --destination pr1961 --name TestApp -p android -p ios -p macos -p windows
No version was specified; fetching available versions...
Using 0.73 because it supports all specified platforms (use --version to specify another version)
Downloading react-native-0.73.6.tgz...
%
```

### DIFF
--- a/scripts/configure.mjs
+++ b/scripts/configure.mjs
@@ -11,6 +11,7 @@ import { cliPlatformIOSVersion } from "./configure-projects.js";
 import {
   getPackageVersion,
   isMain,
+  memo,
   readJSONFile,
   readTextFile,
   toVersionNumber,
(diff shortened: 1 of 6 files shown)


kelset (approved):
locally it works, and changes make sense 👍

PR #2003, opened 2024-04-29
Add an icon to RNTA on windows

acoates-ms:
Replacing blank icon/splash screen images for the windows test apps.


kelset (approved):
love it! but some of the CIs are not that happy with the header image in the readme 😅 maybe try putting it somewhere else in the md?

README.md
```diff
@@ -1,4 +1,5 @@
-# React Native Test App
+
```
tido64: Trim (or just revert this file)

tido64 (commented):
Thanks, I'll have a closer look when I get back at my desk.

README.md
```diff
@@ -1,4 +1,4 @@
-# React Native Test App
+# React Native Test App
```
tido64: You managed to add a space at the end when you removed the newline. 😂

acoates-ms: Cool.. also looks like the CI build cannot handle builds from forks.

tido64: Not sure what you mean. One of the tests are failing because we're including more files than expected: https://github.com/microsoft/react-native-test-app/actions/runs/8882235888/job/24386416320#step:11:1

tido64: Do we need all the extra files?

images/RNTA_full.png
tido64, line 1: The bigger versions of this image is so dirty and full of artifacts, which makes it really hard for PNG to compress them. I don't suppose you have a cleaner version of the full image?

acoates-ms: I just took the image straight from the AI generator.  -- I just added some manual cleanup. -- Not perfect, but certainly helped the file sizes.

tido64: It looks like it lost its transparency in the process.

PR #2023, opened 2024-05-07
fix(apple): fix `embed-manifest` failing to find `app.json`

tido64:
### Description

Fix `embed-manifest` failing to find `app.json`

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [x] visionOS
- [ ] Windows

### Test plan

In https://github.com/react-native-webview/react-native-webview:

1. Bump `react-native-test-app` to 3.7.0
2. Install: `yarn`
3. Copy `rnta/scripts/embed-manifest/swift.mjs` to `webview/node_modules/react-native-test-app/scripts/embed-manifest/swift.mjs`
4. Install Pods: `pod install --project-directory=example/ios`
5. Build: `yarn ios`

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -13,7 +13,6 @@
           - scripts/embed-manifest/kotlin.mjs
           - scripts/embed-manifest/main.mjs
           - scripts/embed-manifest/validate.mjs
-          - scripts/generate-*.mjs
           - scripts/init.mjs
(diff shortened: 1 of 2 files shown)


kelset (approved):
sorry - it took me a while to fix my local simulators.

Was able to test and it works 🎉

PR #2062, opened 2024-06-04
fix(android): set `packageName` to ensure correct app launches

tido64:
### Description

Set `packageName` to ensure correct app launches. This affects Android devs that have several copies of RNTA installed on their device.

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

```
cd example
node --print 'require("./react-native.config.js")'

# Set `android.package` in `app.json` to something other than `com.microsoft.reacttestapp`
node --print 'require("./react-native.config.js")'
```

In both cases, verify that `packageName` is correctly set.

### DIFF
--- a/scripts/configure-projects.js
+++ b/scripts/configure-projects.js
@@ -24,26 +24,32 @@ const {
  */
 
 /**
- * Returns the version number of `@react-native-community/cli-platform-ios`.
- * @param {string} reactNativeDir
+ * Returns the version number of a React Native dependency.
+ * @param {string} packageName
  * @returns {number}
  */
-const cliPlatformIOSVersion = (() => {
-  /** @type {number} */
-  let version;
-  /** @type {(reactNativeDir: string) => number} */
-  return (reactNativeDir) => {
-    if (!version) {
-      version = toVersionNumber(
-        getPackageVersion(
-          "@react-native-community/cli-platform-ios",
-          reactNativeDir
-        )
-      );
+const getRNPackageVersion = (() => {
+  const isTesting = "NODE_TEST_CONTEXT" in process.env;
+  /** @type {Record<string, number>} */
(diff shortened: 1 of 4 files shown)


jbroma:
@tido64 did a quick test, something seems off, this results in the following launch attempt with RNC CLI:

```bash
BUILD SUCCESSFUL in 31s
131 actionable tasks: 15 executed, 116 up-to-date
info Connecting to the development server...
info Starting the app on "emulator-5554"...
Starting: Intent { act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHE
R] cmp=com.testerapp/.com.microsoft.reacttestapp.MainActivity }
Error type 3
Error: Activity class {com.testerapp/com.testerapp.com.microsoft.reacttestapp.MainActivity} does not exist.
```

looks like a problem in CLI maybe? `packageName` get's passed correctly to the `react-native.config.js`

jbroma:
CLI fixed this in https://github.com/react-native-community/cli/pull/2388 - it should work as expected with this change 🎉

tido64:
> CLI fixed this in [react-native-community/cli#2388](https://github.com/react-native-community/cli/pull/2388) - it should work as expected with this change 🎉

Looks like we will have to do a version check before setting `packageName` in that case. We can't regress on older versions of the CLI.

jbroma:
> > CLI fixed this in [react-native-community/cli#2388](https://github.com/react-native-community/cli/pull/2388) - it should work as expected with this change 🎉
> 
> Looks like we will have to do a version check before setting `packageName` in that case. We can't regress on older versions of the CLI.

Agreed. I've verified that the fix for the CLI resolves the issue and together the app launches properly now 🎉

PR #2084, opened 2024-06-12
chore: make `release-notes.mjs` output announcement template

tido64:
### Description

Make `release-notes.mjs` output announcement template

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

```sh
node scripts/release-notes.mjs 3.7.0 3.8.0
```

### DIFF
--- a/android/autolink.mjs
+++ b/android/autolink.mjs
@@ -13,8 +13,7 @@ import {
 
 /**
  * @typedef {import("@react-native-community/cli-types").Config} Config
- * @typedef {import("../scripts/types.js").AndroidDependency} AndroidDependency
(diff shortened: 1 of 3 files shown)


JasonVMo (approved):
The changes look good but am I correct that this only works when running in a git repo? Is there a standard abstraction pattern for repo type?

tido64:
> The changes look good but am I correct that this only works when running in a git repo? Is there a standard abstraction pattern for repo type?

This script is very specific to how RNTA is set up, with Git and semantic-release. And we only use it to generate text that we post in our Discord channel and social media. It's not meant to be consumed by anyone outside the repo.

PR #2089, opened 2024-06-17
bump colored2 to 4.0.0

Jvr2022:


Gemfile.lock
```diff
@@ -9,7 +9,7 @@ GEM
     atomos (0.1.3)
     base64 (0.2.0)
     claide (1.1.0)
-    colored2 (3.1.2)
+    colored2 (4.0.0)
```
tido64, line 12: It looks like `colored2` needs to stay on ~3.1 since it's what `xcodeproj` currently requires: https://github.com/microsoft/react-native-test-app/blob/aa185a8bbade9a9ca5a9f1097fb55c685c88612f/Gemfile.lock#L46-L52

tido64 (changes requested):
Thanks for your contribution! Unfortunately, it doesn't look like we can take this change as is (see reason above).

PR #2126, opened 2024-07-16
fix(android): fix loading of core modules on 0.75 + New Arch

tido64:
### Description

Fix loading of core modules on 0.75 + New Arch: https://github.com/facebook/react-native/commit/7facb32f30d17eb27f870dd531d3dc7ebe4532f6

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

n/a

### DIFF
--- a/android/app/src/main/jni/ComponentsRegistry.cpp
+++ b/android/app/src/main/jni/ComponentsRegistry.cpp
@@ -1,5 +1,7 @@
 #include "ComponentsRegistry.h"
 
+#if !__has_include(<DefaultTurboModuleManagerDelegate.h>)
+
 #include "AutolinkingCompat.h"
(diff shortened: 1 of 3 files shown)


JasonVMo (commented):
Have you gone through and tested the code on downlevel versions as well? The main concern is that we end up referencing modules that don't exist in older versions with the new references (though it looks protected by the right ifdefs)

tido64:
> Have you gone through and tested the code on downlevel versions as well? The main concern is that we end up referencing modules that don't exist in older versions with the new references (though it looks protected by the right ifdefs)

I have and it works. The `#ifdef` is working.

PR #2133, opened 2024-07-19
fix(android): introduce upper Gradle wrapper versions

tido64:
### Description

Gradle 8.9 removes APIs that are still being used by 0.75.

Build log: https://github.com/microsoft/react-native-test-app/actions/runs/9993761517/job/27621913498

Fixed upstream in https://github.com/facebook/react-native/commit/b82d7e100c7a6b03295ed87915f54c477b00c983.

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

React Native versions and the Gradle version ranges that is supported:

- 0.72: [8.1.1, 8.3)
- 0.73: [8.3, 8.9)
- 0.74: [8.6, 8.9)
- 0.75: [8.8, 8.9)
- 0.76 (nightly): 8.9

Test cases should cover these.

To manually test them:
- Use `set-react-version` to set the `react-native` version
- Set the Gradle wrapper version in `example/android/gradle/wrapper/gradle-wrapper.properties`
- Build the Android app under `example/android`: `./gradlew assembleDebug`


android/gradle-wrapper.js
```diff
@@ -62,18 +62,26 @@ function configureGradleWrapper(sourceDir, fs = nodefs) {
       } else if (version >= v(0, 75, 0)) {
         if (gradleVersion < v(8, 8, 0)) {
```
JasonVMo: Feels like this could be refactored to be a bit less manual. Like have a minimum/maximum value set for each RN version, then have the code keep them in bounds. That way if things change you don't have to update the if statements (or add a new one each time). Still this change looks fine.

JasonVMo (approved):
Nit on the if statements but not blocking, just a comment for the future.

PR #2149, opened 2024-07-23
fix(android): update Gradle version check during build

tido64:
### Description

Update Gradle version check during build

### Platforms affected

- [x] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

This must currently be manually tested by modifying `example/android/gradle/wrapper/gradle-wrapper.properties` and running `./gradlew clean` under `example/android`.

### DIFF
--- a/android/test-app-util.gradle
+++ b/android/test-app-util.gradle
@@ -10,39 +10,49 @@ ext.autolinkingInfo = { buildDir ->
 }
 
 ext.checkEnvironment = { rootDir ->
-    def warnGradle = { gradleVersion, gradleVersionRange, reactNativeVersion ->
+    // Keep this in sync with the JS version in `android/gradle-wrapper.js`
+    //
+    // We have two implementations because there are currently two ways to build
+    // the Android app. If built via `@react-native-community/cli`, the JS
+    // script will be run and we can change Gradle version before it is
+    // executed. If it's built with Gradle directly, it's already too late and
+    // the best we can do is to warn the user.
+    def gradleVersions = [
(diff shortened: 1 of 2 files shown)


android/test-app-util.gradle
```diff
@@ -10,39 +10,49 @@ ext.autolinkingInfo = { buildDir ->
...
-    def warnGradle = { gradleVersion, gradleVersionRange, reactNativeVersion ->
+    // Keep this in sync with the JS version in `android/gradle-wrapper.js`
+    //
+    // We have two implementations because there are currently two ways to build
+    // the Android app. If built via `@react-native-community/cli`, the JS
+    // script will be run and we can change Gradle version before it is
+    // executed. If it's built with Gradle directly, it's already too late and
+    // the best we can do is to warn the user.
+    def gradleVersions = [
+      [v(0, 76, 0), [v(8, 9, 0), "8.9"], [Integer.MAX_VALUE, ""]],  //  0.76: [8.9, *)
```
JasonVMo, line 21: Can you pull json files into .gradle files? If so you could put this table in a .json file and pull it into both places so you don't have to keep two code locations in sync. Otherwise looks good.

JasonVMo (approved):

android/test-app-util.gradle
tido64: `v()` calculates the numerical value of a version string, which we can't do in a JSON file. We would have to hard-code them, and if the algorithm changes for some reason (it has changed once, though unlikely again), we would also have to manually update the table. Maybe that's fine?

I'll merge this for now and revisit this later.

PR #2160, opened 2024-07-29
feat(macos): add macOS support for config plugins

tido64:
### Description

I had this branch lying around for about a year or so. Rebased, and brushed up a bit.

### Platforms affected

- [ ] Android
- [ ] iOS
- [x] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

1. Enable config plugins:
    ```diff
    diff --git a/example/app.json b/example/app.json
    index 3b8992ea..8be080e1 100644
    --- a/example/app.json
    +++ b/example/app.json
    @@ -13,6 +13,9 @@
           "presentationStyle": "modal"
         }
       ],
    +  "plugins": [
    +    "./test-plugin.js"
    +  ],
       "resources": {
         "android": [
           "dist/res",
    diff --git a/example/package.json b/example/package.json
    index 35447f2e..b02cc291 100644
    --- a/example/package.json
    +++ b/example/package.json
    @@ -29,6 +29,7 @@
       "devDependencies": {
         "@babel/core": "^7.1.6",
         "@babel/preset-env": "^7.1.6",
    +    "@expo/config-plugins": "^8.0.0",
         "@react-native/babel-preset": "^0.73.19",
         "@react-native/metro-config": "^0.73.3",
         "@rnx-kit/metro-config": "^1.3.15",
    ```
2. Create `example/test-plugin.js`:
    ```js
    const { createRunOncePlugin, } = require("@expo/config-plugins");
    const { mergeContents } = require("@expo/config-plugins/build/utils/generateCode");
    const { macos } = require("react-native-test-app/plugins/index");

    function withCustomAppDelegate(config) {
      return macos.withAppDelegate(config, (config) => {
        config.modResults.contents = mergeContents({
          tag: "test-plugin.js",
          src: config.modResults.contents,
          newSrc: "// added by test-plugin.js",
          anchor: /import Cocoa/,
          offset: 1,
          comment: "//",
        }).contents;
        return config;
      });
    }

    module.exports = createRunOncePlugin(
      withCustomAppDelegate,
      "test-plugin.js",
      "UNVERSIONED"
    );
    ```
3. Under `example` folder, run `pod install --project-directory=macos`
4. Verify that `AppDelegate.swift` was written to:
    ```diff
    diff --git a/macos/ReactTestApp/AppDelegate.swift b/macos/ReactTestApp/AppDelegate.swift
    index 73a7288e..95f936ae 100644
    --- a/macos/ReactTestApp/AppDelegate.swift
    +++ b/macos/ReactTestApp/AppDelegate.swift
    @@ -1,4 +1,7 @@
     import Cocoa
    +// @generated begin test-plugin.js - expo prebuild (DO NOT MODIFY) sync-bf0d29fe87baaf78ed9ffca2f64bacb7c675534f
    +// added by test-plugin.js
    +// @generated end test-plugin.js
     import ReactTestApp_DevSupport

     @main
    ```


scripts/config-plugins/plugins/mod-compiler.mjs
```diff
@@ -13,6 +14,12 @@ export const withDefaultBaseMods = (config, props) => {
     ...props,
     providers: getAndroidModFileProviders(),
   });
+  config = BaseMods.withGeneratedBaseMods(config, {
+    ...props,
+    // @ts-expect-error `macos` is not assignable to type `android | ios`
+    platform: "macos",
```
shirakaba, line 20: Unfortunately, Expo config plugins hard-code the string "ios" all over the place, consequentially searching in the `ios` folder and reading values from `config.ios`, so I'm not sure which plugins will work and which ones won't with this approach.

I opened a big [PR](https://github.com/expo/expo/pull/30660) last weekend to add macOS support to Expo config plugins, but I am not optimistic about it being reviewed anytime soon.

scripts/config-plugins/plugins/cocoaBaseMods.mjs
```diff
@@ -0,0 +1,53 @@
...
+    ),
+    infoPlist: modifyFilePath(expoProviders.infoPlist, "Info.plist"),
+    entitlements: nullProvider,
+    podfile: makeNullProvider({
+      path: "",
+      language: /** @type {const} */ ("rb"),
+      contents: "",
+    }),
+    podfileProperties: makeNullProvider(),
+  };
```
shirakaba, line 44: This is really clever. 👏 

Given RNTA has a predictable structure, a lot of the complexity in the plugins (e.g. searching for all possible locations for an [Info.plist](https://github.com/expo/expo/blob/ea5d3f54cbc93cb7d84f9bb53af62d5f6cebfe84/packages/%40expo/config-plugins/src/ios/Paths.ts#L256) or [Xcodeproj](https://github.com/expo/expo/blob/ea5d3f54cbc93cb7d84f9bb53af62d5f6cebfe84/packages/%40expo/config-plugins/src/ios/Paths.ts#L180)) can be totally sidestepped, as you've done here. And that side-steps a lot of the coupling to `ios`.

I am not sure to what extent this sidesteps *all* of their hard-coding of the `ios` platform, though. But given how your test passes, maybe it's quite a lot. Surely at least one of these will end up referring to `config.ios` or the `ios` folder at some point, though? Do you have an impression of the extent to which Expo config plugins will work?

tido64: As far as I've understood, the providers are just functions for making modifications. They don't have props that are platform specific, but may be pointing to some iOS specific paths as you're suggesting. We should be overwriting all the props here (TypeScript would complain otherwise). Save for the null providers, I think we can say we are complete.

That said, these providers aren't supposed to be platform specific. The `with*` functions (e.g. `withAppDelegate`) are platform specific, and we need a separate set for macOS. That's what I've done here: https://github.com/microsoft/react-native-test-app/blob/622a257b26472a74497d4effae71eb07b627e6da/plugins/macos.js#L35-L79

Which means that for now, we only support these four functions. And you will need to import them from `react-native-test-app/plugins/macos.js`.

scripts/config-plugins/plugins/mod-compiler.mjs
tido64: If I understand this correctly, this part just registers what's available for the `macos` platform. We need to explicitly implement the mods separately, using `withMod`, as in [`plugins/macos.js`](https://github.com/microsoft/react-native-test-app/pull/2160/files#diff-4436c441b99f94f7f3affd3ec2744dc4aff5b30496912e0894fef1b2fe4bff4f).

plugins/index.js
```diff
@@ -34,5 +82,6 @@ function withSceneDelegate(config, action) {
   });
 }
 
+exports.macos = macos;
```
tido64: On second thought, maybe it makes more sense to move these to a separate file so the import becomes `react-native-test-app/plugins/macos`.

Saadnajmi (approved):

PR #2164, opened 2024-08-02
test: refactor test matrix to support more platforms

tido64:
### Description

Refactor test matrix to support more platforms.

More platforms to come in separate PRs.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

```
npm run test:matrix
```

### DIFF
--- a/scripts/test-e2e.mjs
+++ b/scripts/test-e2e.mjs
@@ -29,7 +29,7 @@ export function $(command, ...args) {
  * @param {...string} args
  * @returns {string}
  */
-function $$(command, ...args) {
+export function $$(command, ...args) {
   const { status, stderr, stdout } = spawnSync(command, args, {
     stdio: ["ignore", "pipe", "pipe"],
     encoding: "utf-8",
@@ -50,7 +50,7 @@ function $$(command, ...args) {
 function ensureAppiumAvailable() {
   return new Promise((resolve, reject) => {
     const socket = new Socket();
-    socket.setTimeout(60);
+    socket.setTimeout(60 * 1000);
 
     /** @type {(e: Error) => void} */
     const onError = (e) => {
--- a/scripts/test-matrix.mjs
+++ b/scripts/test-matrix.mjs
@@ -4,80 +4,36 @@
  * Reminder that this script is meant to be runnable without installing
  * dependencies. It can therefore not rely on any external libraries.
  */
-import { spawn, spawnSync } from "node:child_process";
+import { spawn } from "node:child_process";
 import * as fs from "node:fs";
 import { fileURLToPath } from "node:url";
-import { memo, readTextFile } from "./helpers.js";
+import * as util from "node:util";
+import { green, red, yellow } from "./colors.mjs";
+import { readTextFile } from "./helpers.js";
 import { setReactVersion } from "./set-react-version.mjs";
-import { $, test } from "./test-e2e.mjs";
+import {
+  configurePodfile,
+  getIOSSimulatorName,
+  installPods,
+} from "./test-apple.mjs";
+import { $, $$, test } from "./test-e2e.mjs";
 
 /**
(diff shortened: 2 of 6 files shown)


scripts/test-e2e.mjs
```diff
@@ -29,7 +29,7 @@ export function $(command, ...args) {
  * @param {...string} args
  * @returns {string}
  */
-function $$(command, ...args) {
+export function $$(command, ...args) {
```
Saadnajmi, line 32: TIL this is a valid function name.

scripts/test-matrix.mjs
```diff
@@ -4,124 +4,79 @@
  * Reminder that this script is meant to be runnable without installing
  * dependencies. It can therefore not rely on any external libraries.
  */
-import { spawn, spawnSync } from "node:child_process";
+import { spawn } from "node:child_process";
```
Saadnajmi, line 7: I thought we had to use `spawnSync` because of a CVE related to uses of `spawn`?

scripts/test-matrix.mjs
```diff
@@ -4,124 +4,79 @@
...
- * @typedef {ApplePlatform | "android" | "windows"} Platform
+ * @typedef {import("./types.js").BuildConfig} BuildConfig
+ * @typedef {import("./types.js").PlatformConfig} PlatformConfig
+ * @typedef {import("./types.js").TargetPlatform} TargetPlatform
  */
 
-const PACKAGE_MANAGER = "yarn";
-const TAG = "┃";
+const DEFAULT_PLATFORMS = ["android", "ios"];
+const TEST_VARIANTS = /** @type {const} */ (["paper", "fabric"]);
```
Saadnajmi, line 24: Nit: perhaps oldArch / newArch is better?

scripts/test-matrix.mjs
```diff
@@ -4,124 +4,79 @@
...
+  },
+  visionos: {
+    name: "visionOS",
+    engines: ["jsc", "hermes"],
+    configure: configurePodfile,
+    prebuild: installPods,
+  },
+  windows: {
+    name: "Windows",
+    engines: ["hermes"],
```
Saadnajmi, line 64: I guess, no v8?

scripts/test-matrix.mjs
```diff
@@ -163,13 +118,147 @@ function startDevServer(logPath = "metro.server.log") {
...
+      case "android":
+      case "ios":
+        filtered.push(platform);
+        break;
+
+      case "macos":
+      case "visionos":
+      case "windows":
+        log(yellow(`⚠ Unsupported platform: ${platform}`));
+        break;
```
Saadnajmi, line 135: These are the "more platforms coming" I assume?

scripts/test-matrix.mjs
tido64: The issue is only when you execute a `.bat` or `.cmd` (or similar) without a shell on Windows. You just need to make sure you set `shell: true` like here: https://github.com/microsoft/react-native-test-app/blob/2163e571573d7ec34efdcca203171ee4a8d4ffa4/test/android/gradle.ts#L131-L133

scripts/test-matrix.mjs
tido64: Is the V8 path even public?

Saadnajmi: @acoates-ms ?

tido64: For now though, let's keep it like this. Windows will be enabled in the next PR. We can consider V8 then.

acoates-ms: We dont provide support for the V8 path - I wouldn't bother including it.

Saadnajmi (approved):

PR #2188, opened 2024-08-19
fix(macos): add anchor points for `applicationWillFinishLaunching(_:)`

tido64:
### Description

Add anchor points for `applicationWillFinishLaunching(_:)` to make it easier for config plugin authors to target the correct override.

### Platforms affected

- [ ] Android
- [ ] iOS
- [x] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

n/a

### DIFF
--- a/macos/ReactTestApp/AppDelegate.swift
+++ b/macos/ReactTestApp/AppDelegate.swift
@@ -149,6 +149,8 @@ extension AppDelegate {
     }
 
     func applicationWillFinishLaunching(_: Notification) {
+        // applicationWillFinishLaunching(_:) [ENABLE_SINGLE_APP_MODE=0]
+
         if Session.shouldRememberLastComponent {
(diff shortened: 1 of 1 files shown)


shirakaba (approved):
Looks like it'd do the trick, thanks!

Saadnajmi (approved):

Saadnajmi:
@tido64 Long term, should the ifdefs be more granular? AKA, ifdefs inside the method and only one method definition?

tido64:
> @tido64 Long term, should the ifdefs be more granular? AKA, ifdefs inside the method and only one method definition?

For `AppDelegate.swift` in particular, I think we want to create two more files, `AppDelegate+Multi.swift` and `AppDelegate+Single.swift`, and move the appropriate extensions to them. Single app mode requires significantly less logic to start up the app so I think what you're suggesting makes it harder to see what each mode actually implements.

PR #2204, opened 2024-09-02
feat: add :additional_build_settings to options

okwasniewski:
### Description

This change allows us to pass additional build settings to the main generated project:

```ruby
options = {
  :app_build_settings => { 
    'LIBRARY_SEARCH_PATHS' => '$(inherited) "../something"',
  }
}


use_test_app! options 
```

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [x] visionOS
- [ ] Windows

### Test plan

Pass additional build setting.

### DIFF
--- a/ios/pod_helpers.rb
+++ b/ios/pod_helpers.rb
@@ -98,6 +98,16 @@ def try_pod(name, podspec, project_root)
   pod name, :podspec => podspec if File.exist?(File.join(project_root, podspec))
 end
 
+def append_additional_build_settings!(build_settings, settings_to_append)
+  settings_to_append&.each do |setting, value|
+    if build_settings.key?(setting)
+      build_settings[setting] += value
+    else
+      build_settings[setting] = value
+    end
+  end
+end
+
 def use_hermes?(options)
   use_hermes = ENV.fetch('USE_HERMES', nil)
   return use_hermes == '1' unless use_hermes.nil?
--- a/ios/pod_helpers.rb
+++ b/ios/pod_helpers.rb
@@ -98,15 +98,9 @@ def try_pod(name, podspec, project_root)
   pod name, :podspec => podspec if File.exist?(File.join(project_root, podspec))
 end
 
-def append_additional_build_settings!(build_settings, settings_to_append)
+def override_build_settings!(build_settings, settings_to_append)
   settings_to_append&.each do |setting, value|
-    if build_settings.key?(setting)
-      current = build_settings[setting]
-      build_settings[setting] += " #{value}" if current.is_a? String
-      build_settings[setting] += value if current.is_a? Array
-    else
-      build_settings[setting] = value
-    end
+    build_settings[setting] = value
   end
 end
 
--- a/ios/pod_helpers.rb
+++ b/ios/pod_helpers.rb
@@ -101,7 +101,9 @@ end
 def append_additional_build_settings!(build_settings, settings_to_append)
   settings_to_append&.each do |setting, value|
     if build_settings.key?(setting)
-      build_settings[setting] += value
+      current = build_settings[setting]
+      build_settings[setting] += " #{value}" if current.is_a? String
+      build_settings[setting] += value if current.is_a? Array
     else
       build_settings[setting] = value
     end
--- a/ios/test_app.rb
+++ b/ios/test_app.rb
@@ -259,6 +259,8 @@ def make_project!(xcodeproj, project_root, target_platform, options)
     uitests_build_settings[PRODUCT_BUNDLE_IDENTIFIER] = "#{product_bundle_identifier}UITests"
   end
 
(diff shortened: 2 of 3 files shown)


okwasniewski:
Adding some test for this case would be great but I'm not sure how to test options passed to use_test_app, can you give me some clues how to test it? I couldn't find it in the repo

tido64:
> Adding some test for this case would be great but I'm not sure how to test options passed to use_test_app, can you give me some clues how to test it? I couldn't find it in the repo

If we're talking about unit tests, we currently only test end-to-end. If we were to test this properly, I would extract the block where we populate `build_settings` to a separate function, and add tests for it. You can see examples of Ruby tests under `/test`.

ios/test_app.rb
```diff
@@ -259,6 +259,10 @@ def make_project!(xcodeproj, project_root, target_platform, options)
     uitests_build_settings[PRODUCT_BUNDLE_IDENTIFIER] = "#{product_bundle_identifier}UITests"
   end
 
+  options[:additional_build_settings]&.each do |setting, value|
+    build_settings[setting] ||= value
```
tido64: Is this intentional? If `build_settings[setting]` is already set, we don't set it to `value`?

Edit: Oh, is this because it's "additional"?

okwasniewski: Damn, I didn't read the whole doc about `||=` operator. I think a more predictable approach would be to use `+=` and append the value (either string or array) WDYT?

tido64: I think appending makes sense. If we had a `build_settings_overrides`, it would be safe to assume that we overwrite values.

okwasniewski: Changed it to `+=`

test/test_pod_helpers.rb
```diff
@@ -112,4 +112,22 @@ def test_v
...
   end
+
+  def test_append_additional_build_settings
+    build_settings = { 'OTHER_LDFLAGS' => '-l"Pods-TestApp"' }
+
+    append_additional_build_settings!(build_settings, {})
+
+    assert_equal('-l"Pods-TestApp"', build_settings['OTHER_LDFLAGS'])
+
+    append_additional_build_settings!(build_settings, { 'OTHER_LDFLAGS' => ' -ObjC' })
```
tido64: Can you also add a case where `OTHER_LDFLAGS` isn't already set in the destination object?

okwasniewski: Yeah, just spotted this bug, test case added

test/test_pod_helpers.rb
```diff
@@ -112,4 +112,27 @@ def test_v
...
+
+    assert_equal('-l"Pods-TestApp" -ObjC', build_settings['OTHER_LDFLAGS'])
+
+    # Test passing a table
+    build_settings_arr = { 'OTHER_LDFLAGS' => ['$(inherited)', '-l"Pods-TestApp"'] }
+    append_additional_build_settings!(build_settings_arr, { 'OTHER_LDFLAGS' => [' -ObjC'] })
+
+    assert_equal(['$(inherited)', '-l"Pods-TestApp"', ' -ObjC'], build_settings_arr['OTHER_LDFLAGS'])
```
tido64: Can you run `yarn lint:rb -A` so RuboCop doesn't complain about formatting?

ios/pod_helpers.rb
```diff
@@ -98,6 +98,16 @@ def try_pod(name, podspec, project_root)
   pod name, :podspec => podspec if File.exist?(File.join(project_root, podspec))
 end
 
+def append_additional_build_settings!(build_settings, settings_to_append)
+  settings_to_append&.each do |setting, value|
+    if build_settings.key?(setting)
+      build_settings[setting] += value
```
tido64: Can we prefix a space if the value is a string so users don't have to think about it?
```suggestion
      current = build_settings[setting]
      build_settings[setting] += " #{value}" if current.is_a? String
      build_settings[setting] += value if current.is_a? Array
```

test/test_pod_helpers.rb
```diff
@@ -113,27 +113,31 @@ def test_v
...
-                 build_settings_arr['OTHER_LDFLAGS'])
+    assert_equal([' -ObjC'], build_settings_arr['OTHER_LDFLAGS'])
 
     # Test setting a new key
-    append_additional_build_settings!(build_settings, { 'OTHER_CFLAGS' => '-DDEBUG' })
+    override_build_settings!(build_settings, { 'OTHER_CFLAGS' => '-DDEBUG' })
 
     assert_equal('-DDEBUG', build_settings['OTHER_CFLAGS'])
+
+
```
tido64:
```suggestion

```

tido64 (approved):
Thanks for adding this ❤️

PR #2252, opened 2024-09-18
fix(apple): unwrap Node wrappers before writing env files

tido64:
### Description

Unwrap Node wrappers before writing env files

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [x] visionOS
- [ ] Windows

### Test plan

n/a

### DIFF
--- a/ios/node.rb
+++ b/ios/node.rb
@@ -1,0 +1,28 @@
+require('json')
+require_relative('pod_helpers')
+
+def find_node
+  node_bin = `which node`.strip!
+  return node_bin if `file #{node_bin}`.include? 'Mach-O'
+
+  # If a wrapper is found, remove the path from `PATH` and try again. This can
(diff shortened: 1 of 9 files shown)


okwasniewski (approved):

ios/node.rb
```diff
@@ -0,0 +1,30 @@
+require('json')
+require_relative('pod_helpers')
+
+def find_node
+  node_bin = `which node`.strip!
+  return node_bin if `file #{node_bin}`.include? 'Mach-O'
+
+  # If a wrapper is found, remove the path from `PATH` and try again. This can
+  # sometimes happen when `pod install` is run inside a "virtual" environment
+  # like Turborepo.
```
Saadnajmi: Can we add a link here? Turboreoo isn't obvious to me what this is, and I might think it's related to turbo modules.

Saadnajmi (approved):

PR #2256, opened 2024-09-23
fix(windows): fix `run-windows` not launching apps built with New Arch

tido64:
### Description

`run-windows` fails to launch apps built with New Arch because artifacts are output relative to the `.wapproj` under `node_modules/.generated/windows/Win32/bin`, but the command expects them to be under the solution directory.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [x] Windows

### Test plan

```
cd example
yarn install-windows-test-app --use-fabric
yarn windows
```

### DIFF
--- a/windows/Win32/ReactApp.Package.wapproj
+++ b/windows/Win32/ReactApp.Package.wapproj
@@ -13,6 +13,17 @@
   </PropertyGroup>
   <Import Project="$(ReactNativeWindowsDir)\PropertySheets\External\Microsoft.ReactNative.WindowsSdk.Default.props" />
   <PropertyGroup>
+    <!--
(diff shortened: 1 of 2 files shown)


tido64:
cc @Yajur-Grover

Yajur-Grover:
Was able to build and deploy the example app following the steps in the test plan above. Hit with the following error, which I hit previously using RNTA on async-storage with New Architecture:
image

AFAIK, this error is unrelated to the RNTA repo, and is caused by the layoutDirection logic that should be resolved once [this PR](https://github.com/microsoft/react-native-windows/pull/13866) is merged in.

acoates-ms (approved):

PR #2262, opened 2024-09-26
fix(config): make it clear that `configure-test-app` performs a reset

### LINKED ISSUE #2260: Running configure-test-app breaks react-native-builder-bob based projects
Given a project like this:
```
./
./package.json
./android/
./ios/
./example/
./example/ios
./example/android
./example/package.json
```

In a typical react native module development environment (e.g. one initialized by builder bob or hand-rolled) one is able to open `./example/android` in Android Studio and see the WIP package located at `root/android/` and use the Java LSP to work on WIP code. When using react native test app the `root/android` WIP project does not show its files in Android Studio, even after successfully building the project. This prevents me from being able to work in Android Studio on WIP code when the example app is powered by react native test app. 

I don't rule out that this could be a bug but first I want to make sure that this wasn't intentional by the react native test app package

tido64:
### Description

Change the messaging to make it clear that we're doing factory reset.

Resolves #2260.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

```
cd example
yarn configure-test-app
```

Expected output:

```
% yarn configure-test-app
[!] The following files will be reset to their original state:
     .gitignore
     .watchmanconfig
     android/build.gradle
     android/gradle.properties
     android/gradlew
     android/gradlew.bat
     android/settings.gradle
     babel.config.js
     ios/Podfile
     macos/Podfile
     metro.config.js
     react-native.config.js
     windows/.gitignore
[!] The following files will be removed:
     ios/Podfile.lock
     macos/Podfile.lock
[!] Some files will be reset and/or removed: You may have to manually restore any customizations to get the app working again (for more details, see https://github.com/microsoft/react-native-test-app/wiki/Updating#reconfiguringresetting-rnta)
Re-run with --force if you're fine with this.
```


JasonVMo (approved):
The one question here is how often this is run. If this is run all the time then I would warn when we do modify the config. If it is only when configuring a project then the existing behavior makes sense. I'm assuming that this runs as part of project configuration.

tido64:
> The one question here is how often this is run. If this is run all the time then I would warn when we do modify the config. If it is only when configuring a project then the existing behavior makes sense. I'm assuming that this runs as part of project configuration.

On second thought, I think you're right. The warning should only appear when you explicitly run `configure-test-app` in an existing app. Since the command acts as a reset, it's reasonable that it overwrites your configs. I've changed the messaging to be clearer and updated the wiki about this.

shamilovtim:
Consider adding very specific warnings for `react-native-builder-bob` users? Wouldn't hurt and might reduce some thrash / pain for maintainers of both repos

Saadnajmi (approved):

tido64:
> Consider adding very specific warnings for `react-native-builder-bob` users? Wouldn't hurt and might reduce some thrash / pain for maintainers of both repos

I've changed the wording to include "template". I don't think we want to specifically call out `react-native-builder-bob` as there are other templates out there.

PR #2265, opened 2024-09-30
fix: prefilled development team

okwasniewski:
### Description

Hey,

Currently, when I use react-native-test-app, each time I run pod install, I get this `UBF8T346G9` team prefilled in Xcode. I think it should be empty so it doesn't show an error: "Unknown development team," but it shows that you need to add one.

CleanShot 2024-09-30 at 13 12 22@2x

### Platforms affected

- [ ] Android
- [x] iOS
- [ ] macOS
- [x] visionOS
- [ ] Windows

### Test plan

Generate new project and see there is no development team pre-selected


ios/ReactTestApp.xcodeproj/project.pbxproj
```diff
@@ -403,7 +403,7 @@
 				CODE_SIGN_ENTITLEMENTS = App.entitlements;
 				CODE_SIGN_STYLE = Automatic;
 				DEVELOPMENT_ASSET_PATHS = "";
-				DEVELOPMENT_TEAM = UBF8T346G9;
+				DEVELOPMENT_TEAM = "";
```
tido64: Can we remove these lines instead? The macOS project doesn't have them, so it should be fine.

okwasniewski: Yeah makes sense, removed

tido64 (approved):
Thanks ❤️

PR #2266, opened 2024-10-01
chore: migrate build/run commands to `@rnx-kit/cli`

tido64:
### Description

Migrate build/run commands to `@rnx-kit/cli`. As of 0.79, `rnc-cli` automatically installs pods and breaks any setups that do not have CocoaPods installed in a specific way.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

```
npm run test:matrix
```


Saadnajmi (commented):
Is it worth somehow testing that RNTA works with both CLIs?

tido64:
> Is it worth somehow testing that RNTA works with both CLIs?

Yeah, this is mostly to test `@rnx-kit/cli` in more scenarios. I think for this to land, we would have to set up a matrix.

tido64:
> > Is it worth somehow testing that RNTA works with both CLIs?
> 
> Yeah, this is mostly to test `@rnx-kit/cli` in more scenarios. I think for this to land, we would have to set up a matrix.

Scratch that. I think we need to migrate now to have better control over breaking changes.

Saadnajmi (approved):

PR #2267, opened 2024-10-01
fix(apple): improved Node finder

tido64:
### Description

See https://github.com/facebook/react-native/commit/aa35a2171477ec87bc2abb727a31d6fbe1588b54

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [x] visionOS
- [ ] Windows

### Test plan

CI should pass.

### DIFF
--- a/ios/node.rb
+++ b/ios/node.rb
@@ -5,11 +5,7 @@ def find_node
   # If `pod install` is run inside a "virtual" environment like
   # [Yarn](https://yarnpkg.com/), we might find Node wrappers instead of the
   # actual binary.
-  paths = `type --all --path node`.split("\n")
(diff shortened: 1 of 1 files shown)


tido64:
cc @okwasniewski

Saadnajmi (approved):

atlj:
@tido64 thanks for this. Looks like this also fixed an issue with [asdf](https://asdf-vm.com/). With the way `asdf` treats the `node` shim, calling `file` on it returns `Bourne-Again shell script text executable, ASCII text`. Which doesn't include `Mach-O` therefore raising an error.

Here's an example from bob https://github.com/callstack/react-native-builder-bob/issues/639

PR #2281, opened 2024-10-15
chore(dependency): update `react-native-windows` version

Yajur-Grover:
### Description
Bump minimum version of RNW to 0.75.5 (latest) to include the following PR that was backported, to allow apps using RNTA on New Architecture to launch with no issue: https://github.com/microsoft/react-native-windows/pull/13866

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [x] Windows

### Test plan

Example app screenshots:

Old Architecture:
image

New Architecture:
image

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -24,7 +24,7 @@
     "react": "^18.2.0",
     "react-native": "^0.75.0",
     "react-native-macos": "^0.75.2",
-    "react-native-windows": "^0.75.0"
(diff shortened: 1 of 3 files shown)


tido64:
It looks like build is failing because of missing NuGet artifacts. Is this PR really necessary? Users can bump to .5 today without this.

PR #2407, opened 2025-03-05
chore: disable automatic Pod install

tido64:
### Description

Disable automatic Pod install

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

CI should pass.

### DIFF
--- a/example/react-native.config.js
+++ b/example/react-native.config.js
@@ -6,6 +6,9 @@ const project = (() => {
         sourceDir: "android",
       },
       ios: {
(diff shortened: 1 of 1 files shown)


tido64:
Looks like I can't disable it without affecting newly generated projects so I'm just going to opt to use `rnx-cli` instead: https://github.com/microsoft/react-native-test-app/pull/2266

Issue #2411, opened 2025-03-12
Apple: Prepare for the migration away from CocoaPods


tido64:
React Native is migrating away from CocoaPods. We currently have a lot of Ruby code that are invoked via CocoaPods. This task is to port any independent scripts to JS and eventually hook into a hypothetical prebuild step. Any change we make must be backwards compatible.

- [x] `ios/assets_catalog.rb` → https://github.com/microsoft/react-native-test-app/pull/2412
- [x] `ios/entitlements.rb` → https://github.com/microsoft/react-native-test-app/pull/2420
- [x] `ios/info_plist.rb` → https://github.com/microsoft/react-native-test-app/pull/2421
- [x] `ios/node.rb` → https://github.com/microsoft/react-native-test-app/pull/2427
- [ ] `ios/pod_helpers.rb`
  - [x] https://github.com/microsoft/react-native-test-app/pull/2422
  - [x] https://github.com/microsoft/react-native-test-app/pull/2435
- [x] `ios/privacy_manifest.rb` → https://github.com/microsoft/react-native-test-app/pull/2419
- [ ] `ios/test_app.rb`
  - [x] https://github.com/microsoft/react-native-test-app/pull/2422
  - [x]  https://github.com/microsoft/react-native-test-app/pull/2427
  - [x] https://github.com/microsoft/react-native-test-app/pull/2442
  - [x] https://github.com/microsoft/react-native-test-app/pull/2443
- [ ] `ios/use_react_native-0.70.rb`
- [ ] `ios/use_react_native-0.71.rb`
- [x] `ios/xcode.rb` → https://github.com/microsoft/react-native-test-app/pull/2412
- [x] `test/test_node.rb` → https://github.com/microsoft/react-native-test-app/pull/2427
- [ ] `test/test_pod_helpers.rb`
  - [x] https://github.com/microsoft/react-native-test-app/pull/2422
  - [x] https://github.com/microsoft/react-native-test-app/pull/2435
- [x] `test/test_test_app.rb` → https://github.com/microsoft/react-native-test-app/pull/2425
- [x] `test/test_xcode.rb` → https://github.com/microsoft/react-native-test-app/pull/2412

tido64:
[...]

PR #2412, opened 2025-03-12
refactor(apple): port some CocoaPods scripts to JS

tido64:
### Description

This commit adds JS ports of:

- `ios/assets_catalog.rb`
- `ios/xcode.rb`

See #2411.

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [x] visionOS
- [ ] Windows

### Test plan

Tests should pass.


Saadnajmi (approved):
I take it these ports aren't going to be called from cocoapods, so for now any updates will need to be ported to both versions of each file?

tido64:
> I take it these ports aren't going to be called from cocoapods, so for now any updates will need to be ported to both versions of each file?

They will be called from CocoaPods eventually. I want to make the port first, then make the switch. We should ideally only call into Node once and get all the info we need for CocoaPods specific things. Until then, updates will need to be done in both files.

PR #2415, opened 2025-03-17
fix: Show New Architecture experimental warning only for older React Native versions

szymonrybczak:
### Description

New Architecture is stable already :)

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [x] visionOS
- [ ] Windows

### Test plan

n/a


tido64 (commented):
Thanks! Can we condition this on the version instead? Say anything below 0.76 should have this warning?

szymonrybczak:
> Can we condition this on the version instead? Say anything below 0.76 should have this warning?

Updated ✅

tido64 (approved):

PR #2428, opened 2025-03-26
refactor(apple): move host and autolinking resolution to JS

tido64:
### Description

Move host and autolinking resolution to JS

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [x] visionOS
- [ ] Windows

### Test plan

```
npm run set-react-version 0.74
yarn clean
yarn
cd example
rm ios/Podfile.lock
```

### DIFF
--- a/test/ios/app.test.ts
+++ b/test/ios/app.test.ts
@@ -23,12 +23,23 @@ describe("generateProject()", macosOnly, () => {
     return generateProjectActual(projectRoot, platform, options, mockfs);
   }
 
-  function makeMockProject(overrides?: Record<string, unknown>) {
+  function makeMockProject(
+    overrides?: Record<string, unknown>,
+    reactNativeVersion = "1000.0.0"
+  ) {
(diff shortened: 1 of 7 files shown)


JasonVMo (approved):

test/ios/app.test.ts
```diff
@@ -376,8 +417,11 @@ const PROJECT_FILES = {
           PRODUCT_VERSION: "1.0",
           USER_HEADER_SEARCH_PATHS: ["/~/node_modules/.generated"],
         },
+        communityAutolinkingScriptPath: undefined,
```
Saadnajmi, line 420: Are these meant to be undefined?

tido64: Yes, they are only required for older versions. See inline comment: https://github.com/microsoft/react-native-test-app/blob/627318d3b95b71fadc836779a41d8522b62beb71/ios/app.mjs#L75-L81

PR #2429, opened 2025-03-26
fix: look for other formats of React Native config

szymonrybczak:
### Description

Inside RNC CLI we added support for TypeScript configs, in this PR I adjusted validation to also include `.ts`, `.cjs`  and `.mjs`.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

Convert JS config to TypeScript and `configureProject` function shouldn't fail.

PR #2430, opened 2025-03-27
fix: support more extensions when looking for `react-native.config.*`

tido64:
### Description

Support more extensions when looking for `react-native.config.*`

See also #2429

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

n/a

### DIFF
--- a/.github/actions/setup-toolchain/action.yml
+++ b/.github/actions/setup-toolchain/action.yml
@@ -13,6 +13,9 @@ inputs:
   java-version:
     description: Desired Java version
     default: "17"
+  node-version:
+    description: Desired Node version
+    default: "22"
   xcode-developer-dir:
     description: Set the path for the active Xcode developer directory
 runs:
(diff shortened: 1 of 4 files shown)


tido64:
cc @szymonrybczak

szymonrybczak (approved):
😮 Really interesting! 

The thing is that if someone will call `configureProjects` function not directly in `react-native.config.*` file but through some helper in other file the value inside stacktrace will be wrong, but there's always fallback to default search.

tido64:
> The thing is that if someone will call `configureProjects` function not directly in `react-native.config.*` file but through some helper in other file the value inside stacktrace will be wrong, but there's always fallback to default search.

Hmm... we should probably allow passing in the project root to avoid searching altogether. If we search the whole stack, we could end up in weird places outside the project. None of the alternatives are robust.

szymonrybczak:
@tido64 you can also leverage `root` property from `react-native config` command which is based on nearest `package.json` IIRC

Saadnajmi (approved):

tido64:
> @tido64 you can also leverage `root` property from `react-native config` command which is based on nearest `package.json` IIRC

I believe that would cause a loop. `react-native config` reads a `react-native.config.js` that calls `configureProject()` which calls `react-native config`.

PR #2444, opened 2025-04-09
fix(apple): always enable Hermes from 0.80+

### LINKED ISSUE #2436: Apple: JavaScriptCore has been extracted from react-native
https://github.com/react-native-community/javascriptcore

tido64:
### Description

With JSC no longer being a part of core from 0.80, we will also drop support for it.

Resolves #2436.

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [x] visionOS
- [ ] Windows

### Test plan

n/a


ios/features.mjs
```diff
@@ -60,16 +60,24 @@ export function isBridgelessEnabled(reactNativeVersion, options) {
...
+      typeof useHermes === "string"
+        ? useHermes === "1"
+        : options["hermesEnabled"] === true;
 
-  // Hermes prebuilds for visionOS was introduced in 0.76
-  if (enabled && platform === "visionos" && reactNativeVersion < v(0, 76, 0)) {
-    return "from-source";
+    // Hermes prebuilds for visionOS was introduced in 0.76
+    if (
+      enabled &&
```
JasonVMo, line 72: It might be better to centralize these types of settings into capabilities. Maybe a function that returns an object with various capabilities given a RN version and process.env settings. It ends up being easier to manage. Doesn't need to be done now but something for you to consider.

JasonVMo (approved):

ios/features.mjs
tido64: That's what this function is used for. See https://github.com/microsoft/react-native-test-app/blob/trunk/ios/app.mjs#L236-L255 and the tests in https://github.com/microsoft/react-native-test-app/blob/trunk/test/ios/app.test.ts

PR #2470, opened 2025-05-08
test: starting with 0.80, JSC is no longer available

tido64:
### Description

Starting with 0.80, JSC is no longer available so we can't test it either.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

```sh
npm run test:matrix 0.80 -- --ios
```

### DIFF
--- a/example/test/specs/app.spec.mjs
+++ b/example/test/specs/app.spec.mjs
@@ -1,5 +1,5 @@
 // @ts-check
-import { equal } from "node:assert/strict";
+import { equal, match } from "node:assert/strict";
 import { after, before, describe, it } from "node:test";
 import { remote } from "webdriverio";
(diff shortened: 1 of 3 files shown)


Saadnajmi (approved):

matthargett:
you can use @okwasniewski JavaScriptCore package, if you still have users that need/want it. https://github.com/react-native-community/javascriptcore

tido64:
> you can use @okwasniewski JavaScriptCore package, if you still have users that need/want it. https://github.com/react-native-community/javascriptcore

Thanks for the heads up! For RNTA, we decided it makes sense to drop JSC to reduce the test matrix/overhead for 0.80+ (see also https://github.com/microsoft/react-native-test-app/pull/2444)

PR #2478, opened 2025-05-20
chore: migrate from `semantic-release` to `nx`

tido64:
### Description

Migrate from `semantic-release` to `nx` to better support monorepos.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

n/a


package.json
```diff
@@ -124,10 +124,13 @@
     "@babel/preset-env": "^7.20.0",
     "@expo/config-plugins": "^9.0.0",
     "@microsoft/eslint-plugin-sdl": "^1.0.0",
+    "@nx/js": "^21.0.0",
```
Saadnajmi, line 127: 👍🏽

.github/workflows/build.yml
```diff
@@ -652,12 +652,18 @@ jobs:
       - name: Verify tarball content
         run: |
           yarn test test/pack.test.ts
+      - name: Release (dry run)
+        if: ${{ github.ref != 'refs/heads/trunk' }}
+        run: |
+          yarn nx release --dry-run
```
Saadnajmi, line 658: The dry run needs a GitHub auth token to read GitHub releases if you make those, I believe.

Saadnajmi (approved):

.github/workflows/build.yml
tido64: I think this should cover it:
image

PR #2651, opened 2026-01-26
chore: bump `react-native` to 0.81

tido64:
### Description

Bumps `react-native` to 0.81 

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

CI should pass.


acoates-ms (approved):

.github/workflows/build.yml
```diff
@@ -450,7 +450,7 @@ jobs:
     name: "visionOS"
     permissions: {}
     runs-on: macos-15
-    if: ${{ github.event_name != 'schedule' }}
+    if: false # visionOS temporarily disabled until releases are resumed
```
Saadnajmi, line 453: Perhaps it should switch to use the rnm version if we want it..

Saadnajmi (approved):

.github/workflows/build.yml
tido64: If anyone cares enough for visionOS, I guess we can. But separately from this PR.

PR #2660, opened 2026-02-03
ci: disable npm cache to ensure build consistency

tido64:
### Description

Ensure `react-native-windows` path can be found

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [x] Windows

### Test plan

CI should pass

### DIFF
--- a/packages/app/windows/Win32/ReactApp.Package.wapproj
+++ b/packages/app/windows/Win32/ReactApp.Package.wapproj
@@ -10,7 +10,7 @@
     <BackgroundTaskDebugEngines>NativeOnly</BackgroundTaskDebugEngines>
   </PropertyGroup>
   <PropertyGroup Label="ReactNativeWindowsProps">
-    <ReactNativeWindowsDir Condition="'$(ReactNativeWindowsDir)'==''">$([MSBuild]::GetDirectoryNameOfFileAbove($(SolutionDir), 'node_modules\react-native-windows\package.json'))\node_modules\react-native-windows\</ReactNativeWindowsDir>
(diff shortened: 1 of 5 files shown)


packages/app/windows/Win32/ReactApp.Package.wapproj
```diff
@@ -10,7 +10,7 @@
     <BackgroundTaskDebugEngines>NativeOnly</BackgroundTaskDebugEngines>
   </PropertyGroup>
   <PropertyGroup Label="ReactNativeWindowsProps">
-    <ReactNativeWindowsDir Condition="'$(ReactNativeWindowsDir)'==''">$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), 'node_modules\react-native-windows\package.json'))\node_modules\react-native-windows\</ReactNativeWindowsDir>
+    <ReactNativeWindowsDir Condition="'$(ReactNativeWindowsDir)'==''">$([MSBuild]::GetDirectoryNameOfFileAbove($(SolutionDir), 'node_modules\react-native-windows\package.json'))\node_modules\react-native-windows\</ReactNativeWindowsDir>
```
acoates-ms: My understanding is that SolutionDir isn't a great thing to rely on, because its possible to attempt to build the project without building the solution, in which case the SolutionDir wouldnt be set.  (I dont know off hand how to trigger such a build).

But if this fixes something I'm ok with it, since I think its somewhat of an edge case.

acoates-ms (approved):

packages/app/windows/Win32/ReactApp.Package.wapproj
tido64: On second thought, I was thinking of potential hoisting issues but that should be fine in this case because we create a copy of it in a known place. I'll revert this.

JasonVMo (approved):

PR #2685, opened 2026-02-25
chore: run `oxfmt` from workspace root

tido64:
### Description

Run `oxfmt` from workspace root

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

n/a

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
-  "private": true,
   "name": "@microsoft/root",
   "version": "0.0.1-dev",
+  "private": true,
   "homepage": "https://github.com/microsoft/react-native-test-app",
   "license": "MIT",
   "author": {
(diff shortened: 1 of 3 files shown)


packages/app/package.json
```diff
@@ -20,6 +20,17 @@
...
   },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/microsoft/react-native-test-app.git",
+    "directory": "packages/app"
+  },
+  "bin": {
+    "init": "scripts/init.mjs",
+    "init-test-app": "scripts/init.mjs",
+    "configure-test-app": "scripts/configure.mjs",
```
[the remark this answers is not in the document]
tido64: The order matters because `bunx` will always choose the first one. See #2418.

package.json
```diff
@@ -12,10 +12,15 @@
...
+  "workspaces": [
+    ".",
+    "packages/app",
+    "packages/app/example"
+  ],
   "scripts": {
     "format": "nx run-many --target format:c,format:js,format:swift",
     "lint": "nx run-many --target lint:js,lint:kt,lint:rb,lint:swift",
-    "format:js": "oxfmt $(git ls-files '.github/*.json' '.github/*.yml' '.yarn/plugins/*' '.yarnrc.yml' 'CONTRIBUTING.md' 'README.md' 'nx.json')",
+    "format:js": "oxfmt '**/README.md' '**/package.json' '*.{cjs,cts,js,mjs,mts,ts,tsx,yml}' '.github/**/*.json' 'CONTRIBUTING.md' 'nx.json' '!packages/app/package.json'",
```
Saadnajmi, line 23: I guess you don't want to invoke via a script or nx because the slowdown is significant?

Saadnajmi (approved):

package.json
Saadnajmi: Would the leading ** be a perf problem? And is there not a config you can list the files in?

Saadnajmi (approved):

package.json
tido64: > I guess you don't want to invoke via a script or nx because the slowdown is significant?

This can be invoked via nx. See the `format` command.

> Would the leading ** be a perf problem? And is there not a config you can list the files in?

The ** is required to match files in subdirectories. I haven't been able to find any configs to list files.

PR #2734, opened 2026-04-10
chore: split example apps

tido64:
### Description

Split example apps to allow running React Native at different versions e.g.:

- React Native 0.85
- React Native macOS 0.81
- React Native Windows 0.82

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

CI should pass.


packages/app/example/package.json
```diff
@@ -20,10 +20,10 @@
   },
   "dependencies": {
     "@react-native-webapis/web-storage": "^0.4.5",
-    "react": "19.1.0",
-    "react-native": "^0.81.0",
-    "react-native-macos": "^0.81.1",
-    "react-native-safe-area-context": "^5.5.2",
+    "react": "19.1.4",
+    "react-native": "^0.81.6",
+    "react-native-macos": "^0.81.0",
```
JasonVMo, line 25: Curious as to why not a later version of react-native-macos? I thought 81.0 was still on a different version of react as well.

JasonVMo (approved):
Generally looks good. I left a comment about the macos version but given I saw it in two places, including downleveling it seems intentional.

packages/app/example/package.json
tido64: The down-leveling was because of align-deps, but you can see in `yarn.lock` that it resolved to a later version. It doesn't really make a difference.

PR #2780, opened 2026-05-21
fix(windows): simplify app initialization with `ReactNativeWin32App`

tido64:
### Description

Simplify app initialization with `ReactNativeWin32App`

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [x] Windows

### Test plan

```
cd packages/example-windows
node ../app/windows/app.mjs
yarn windows
```

### Screenshots

| New Arch | Old Arch |
| :-: | :-: |
| image | image |

### DIFF
--- a/packages/app/windows/Shared/ReactInstance.cpp
+++ b/packages/app/windows/Shared/ReactInstance.cpp
@@ -48,9 +48,14 @@ namespace
 
     std::filesystem::path GetBundleRootPath()
     {
+#if USE_FABRIC
         WCHAR modulePath[MAX_PATH];
-        GetModuleFileNameW(NULL, modulePath, MAX_PATH);
+        GetModuleFileNameW(nullptr, modulePath, MAX_PATH);
+        PathCchRemoveFileSpec(modulePath, MAX_PATH);
         return std::filesystem::path{modulePath}.replace_filename(L"Bundle") / L"";
+#else   // USE_FABRIC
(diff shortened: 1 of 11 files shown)


packages/app/windows/Shared/ReactInstance.cpp
```diff
@@ -104,74 +111,31 @@ std::vector<std::wstring_view> const ReactTestApp::JSBundleNames = {
...
-bool ReactInstance::LoadJSBundleFrom(JSBundleSource source)
+bool ReactInstance::LoadJSBundleFrom(JSBundleSource source, bool reloadHost)
 {
     source_ = source;
 
     auto instanceSettings = reactNativeHost_.InstanceSettings();
     switch (source) {
         case JSBundleSource::DevServer:
-            instanceSettings.JavaScriptBundleFile(L"index");
+            instanceSettings.DebugBundlePath(L"index");
```
acoates-ms: You should just set DebugBundlePath and JavaScriptBundleFile always.  The only think you need to change for lading from DevServer vs Embedded is UseFastRefresh.

tido64: Isn't that wrong? `UseFastRefresh` can be disabled in dev mode if you wish to manually reload. Using it to determine whether to use the embedded JS bundle seems wrong.

acoates-ms (approved):

PR #2796, opened 2026-06-03
ci: use `macos-26` runners

tido64:
### Description

Trying out `macos-26` runners

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

n/a


tido64:
macOS builds cannot use `macos-26` runners because we're on an older version of `fmt` that fails to build with Xcode 26.

JasonVMo (approved):

PR #2819, opened 2026-06-16
feat: load template config from external packages

### LINKED ISSUE #2211: Support out-of-tree platforms
## Motivation

It's currently hard to add new platforms to RNTA without contributing directly in the repository.

## Solution

Move platform folders to separate packages and design a plugin interface so that platforms can be sourced from third-party packages.

## Work

- [x] https://github.com/microsoft/react-native-test-app/issues/1797
  - This work is mostly to reduce the surface area and avoid conflicts
- [ ] Convert the repository to a monorepo
  - [x] https://github.com/microsoft/react-native-test-app/pull/2478 (because it does not support monorepos)
  - [x] https://github.com/microsoft/react-native-test-app/pull/2553
  - [ ] Reconfigure Nx Release to [create unique tags for each project](https://nx.dev/recipes/nx-release/release-projects-independently#create-a-git-tag-for-each-project)
- [ ] Move template configuration to their respective platform folders
- [ ] Move platform folders to separate packages
- [ ] Design plugin interface for out-of-tree platform packages

tido64:
### Description

Allows initializing and configuring existing apps to support additional platforms provided by external packages.

Part of #2211.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

Configure example app to add web support:

```sh
cd packages/app/example
node ../scripts/configure.mjs -p web --force
```

Verify that:

- A `build:web` script was added to `package.json`
- `react-native-web` was added under `dependencies`
- `web/index.html` was created

### DIFF
--- a/packages/app/scripts/configure.mjs
+++ b/packages/app/scripts/configure.mjs
@@ -11,17 +11,13 @@
  *   PlatformPackage,
  * } from "./types.js";
  */
+import { loadContext } from "@rnx-kit/tools-react-native/context";
 import * as nodefs from "node:fs";
 import { createRequire } from "node:module";
 import * as path from "node:path";
 import { URL, fileURLToPath } from "node:url";
 import semverCoerce from "semver/functions/coerce.js";
 import semverSatisfies from "semver/functions/satisfies.js";
-import { getTemplate as getAndroidTemplate } from "../android/template.config.mjs";
-import { getTemplate as getIOSTemplate } from "../ios/template.config.mjs";
-import { getTemplate as getMacOSTemplate } from "../macos/template.config.mjs";
-import { getTemplate as getVisionOSTemplate } from "../visionos/template.config.mjs";
-import { getTemplate as getWindowsTemplate } from "../windows/template.config.mjs";
 import {
   getPackageVersion,
   isMain,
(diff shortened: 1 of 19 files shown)


packages/app/scripts/configure.mjs
```diff
@@ -334,13 +372,17 @@ export function gatherConfig(params, disableCache = false) {
           oldFiles: platformConfig.oldFiles.map((file) => {
             return path.join(platform, file);
           }),
+          dependencies: {
+            ...dependencies,
+            ...platformConfig.dependencies,
```
JasonVMo, line 377: Should the order of these two be reversed? In other words, should the platform config be a baseline with the dependencies taking precedence?

JasonVMo (approved):
Generally looks good, one question for follow up but approving in case you think the ordering is correct as it stands.

packages/app/scripts/configure.mjs
tido64: `dependencies` comes from RNTA itself

https://github.com/microsoft/react-native-test-app/blob/ee075b3e49a63475cf6885efe80f3425d756fb61/packages/app/package.json#L148-L169

We want to be able to override them.

PR #2835, opened 2026-07-01
chore: set up Knip

tido64:
### Description

Set up Knip and fix issues

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

CI should pass


.github/workflows/build.yml
```diff
@@ -249,7 +252,7 @@ jobs:
       - name: react-native run-ios
         if: ${{ steps.affected.outputs.ios != '' }}
         run: |
-          ../scripts/react-native.mts run-ios
+          ../scripts/react-native.ts run-ios
```
Saadnajmi, line 255: Why did this change?

Saadnajmi (approved):

.github/workflows/build.yml
tido64: Because tools like oxfmt use the `type` field in `package.json` to determine whether files are ESM and complain if they're not. I've finally set root to be ESM so we don't need to use the `.mjs`/`.mts` extension any longer.

PR #2845, opened 2026-07-10
chore: bump `@rnx-kit/react-native-host` to 0.5.21

tido64:
### Description

Since we're now using `ReactNativeFeatureFlagsOverridesOSSStable` when it's available, we no longer need to keep feature flags in sync with upstream.

### Platforms affected

- [ ] Android
- [x] iOS
- [x] macOS
- [x] visionOS
- [ ] Windows

### Test plan

n/a


tido64:
Oops: https://github.com/microsoft/rnx-kit/pull/4245

Saadnajmi (approved):

PR #2892, opened 2026-07-31
refactor: use built-in `Error.isError`

tido64:
### Description

Use built-in [`Error.isError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/isError) instead of `instanceof`.

Note that this was recently introduced in Node 24.3. We cannot use it in public-facing code.

### Platforms affected

- [ ] Android
- [ ] iOS
- [ ] macOS
- [ ] visionOS
- [ ] Windows

### Test plan

n/a

### DIFF
--- a/packages/app/scripts/testing/test-matrix.mts
+++ b/packages/app/scripts/testing/test-matrix.mts
@@ -404,7 +404,7 @@ if (platforms.length === 0) {
         process.exitCode = e;
       } else {
         process.exitCode = 1;
-        showBanner(`❌ ${(e instanceof Error && e.stack) || e}`);
+        showBanner(`❌ ${(Error.isError(e) && e.stack) || e}`);
       }
     });
 }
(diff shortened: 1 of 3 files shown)


JasonVMo (commented):
Should we create a wrapper abstraction like?:
```ts
export function isError(e: unknown): e is Error {
  return e != null && (e as { isError?: unknown }).isError || e instanceof Error);
}
```
Generally we can catch the truthy value of isError which will match the new behavior but fall through to the older instance of behavior. It seems like the isError check effectively widens the type.

tido64:
> Should we create a wrapper abstraction like?:

For this scenario specifically, we can use `Error.isError` because they're all for internal use only. There is exactly one instance, in public-facing code, where we use an abstraction: https://github.com/microsoft/react-native-test-app/blob/a924b3a0e5102edc92e1113daeb2079ed159cb5d/packages/app/windows/project.mjs#L36-L38

Which can benefit from being replaced with `Error.isError` (because we don't care what kind of error it is). But we can change it when we bump minimum Node version.

Saadnajmi (approved):

